PR #16, opened 2016-10-19
Editing or adding a site, folders are not sorted in alphabetical order

blindly:
Folders should be sorted in alphabetical order when adding or editing a
site.

Code not tested but was used in other parts of the site.


kspearrin:
I don't think this will work unless you implement https://github.com/bitwarden/web/blob/master/src/Web/wwwroot/app/vault/vaultController.js#L61

blindly:
Added folderSort function

PR #17, opened 2016-10-19
Adding permanent add site button to each folder

blindly:
When a folder has no sites, there's a convenient 'add site' button. It would be nice if this was permanent. 

It's tedious to have to select a folder each time, causing at least in my case to forget about the folder selection since it wasn't used on the first site that was added (due to the convenient add site button).

### DIFF
--- a/src/Web/wwwroot/app/vault/views/vault.html
+++ b/src/Web/wwwroot/app/vault/views/vault.html
@@ -12,7 +12,7 @@
         <div class="box-header with-border">
             <h3 class="box-title"><i class="fa fa-folder-open"></i> {{folder.name}}</h3>
             <div class="box-tools pull-right">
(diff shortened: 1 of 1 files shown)


src/Web/wwwroot/app/vault/views/vault.html
```diff
@@ -12,6 +12,9 @@
         <div class="box-header with-border">
             <h3 class="box-title"><i class="fa fa-folder-open"></i> {{folder.name}}</h3>
             <div class="box-tools pull-right">
+                <button type="button" class="btn btn-default btn-flat" ng-click="addSite(folder)" uib-tooltip="Add Site">
```
kspearrin: This is incorrect btn styles. Should be `btn-box-tool like the one below it

PR #20, opened 2016-10-24
Autofocus on login page and 2fa

blindly:
Autofocuses on the username field.
Autofocuses on the 2fa code field. The user shouldn't have to click on the input box to enter the 2fa code. It should be autofocused.

### DIFF
--- a/src/Web/wwwroot/app/accounts/views/accountsLoginInfo.html
+++ b/src/Web/wwwroot/app/accounts/views/accountsLoginInfo.html
@@ -8,7 +8,7 @@
     </div>
     <div class="form-group has-feedback" show-errors>
         <label for="email" class="sr-only">Email</label>
-        <input type="email" id="email" name="Email" class="form-control" placeholder="Email" ng-model="model.email"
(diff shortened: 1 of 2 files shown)


src/Web/wwwroot/app/accounts/views/accountsLoginInfo.html
```diff
@@ -8,7 +8,7 @@
     </div>
     <div class="form-group has-feedback" show-errors>
         <label for="email" class="sr-only">Email</label>
-        <input type="email" id="email" name="Email" class="form-control" placeholder="Email" ng-model="model.email"
+        <input type="email" id="email" name="Email" class="form-control" placeholder="Email" ng-model="model.email" autofocus
```
kspearrin: Does this work? I can't fond anything about an `autofocus` directive being part of angularjs without writing something custom...

blindly: Should work. Most likely requires this as well on the controller.
https://gist.github.com/mlynch/dd407b93ed288d499778#file-autofocus-js

Either that or this as a global. 
https://gist.github.com/kirkstrobeck/599664399dbc23968741

PR #23, opened 2016-10-27
Fix collapsed sidebar and its links

vueu:
So I ran into an issue. For example, when you want to use a sub-option of an option on the collapsed sidebar, the controller isn't in use because you are not on the specific route that uses that controller. 

To work around this, I created a custom function that took two params. The intended state, and the function you'd like to be called (which was addSite, addFolder, changeEmail, changePassword, ...) Since controllers are called upon via states, say for example you would like to access a settings function (such as changePassword.) Well you can't unless you are on the settings state.

This function looked to see if you were on the correct state, and if you were, it would call the function intended (again, changePassword, changeEmail, etc.) If you were not on the correct state, meaning the controller corresponding to the action (function) you would like to call is not available, then the state would be changed to the proper one, and then the action (function) you intended to use would be called!

I went about it with the `.then()` function on `$state.go()`, like this:

```
$state.go('example.state').then(function() {
    // ...
});
```

**Edit:** Just realized I still have a silly console.log()'s in my code. I'll remove it, and just know I use them for debugging quite frequently so it was a mistake to leave it in.

### DIFF
--- a/src/Web/wwwroot/app/global/mainController.js
+++ b/src/Web/wwwroot/app/global/mainController.js
@@ -43,7 +43,6 @@ angular
         };
 
         $scope.toggleSideNav = function() {
-            console.log('hi');
             $scope.$broadcast('settingsToggleSideNav');
         };
 


src/Web/wwwroot/app/settings.js
```diff
@@ -1,2 +1,2 @@
 angular.module("bit")
-.constant("appSettings", {"rememberedEmailCookieName":"bit.rememberedEmail","version":"1.2.2","environment":"Development","apiUri":"http://localhost:4000"});
+.constant("appSettings", {"rememberedEmailCookieName":"bit.rememberedEmail","version":"1.3.0","environment":"Development","apiUri":"http://localhost:4000"});
```
bin0: Sorry, wasn't sure if I was supposed to build or not but I did have to change it to production because I did not have a copy of the server on my computer, and I guess it tampered with this file.

src/Web/wwwroot/app/global/mainController.js
```diff
@@ -36,6 +32,20 @@ angular
...
+              callback();
+            } else {
+                $state.go(state).then(function() {
+                  callback();
+                });
+            }
+        };
+
+        $scope.toggleSideNav = function() {
+            $scope.$broadcast('settingsToggleSideNav');
```
bin0: Not sure, was trying to blend it in with some steps you had already taken. I'll switch it to that way.

src/Web/wwwroot/app/services/sideNavService.js
```diff
@@ -0,0 +1,24 @@
+angular
+    .module('bit.services')
+
+    .factory('sideNavService', function () {
+        var _service = {};
+
+        _service.toggleSideNav = function () {
+            if (localStorage.showSideNav == "false") {
```
bin0: Cool, and sorry. I use this frequently however I wasn't sure if it was included in the project or not, and I couldn't find it but will do.

kspearrin:
@bin0 are you still working on this?

bin0:
@kspearrin Sorry, I'll finish it tonight. Been busy!

PR #109, opened 2017-10-12
Add 'Deselect All' option once 'Select All' is clicked in folder

ysshah:
Just thought this would be a useful feature. Let me know if you'd like me to make any changes.


ysshah:
I just realized that this might be the intended functionality, in case the user deselects a few checkboxes and wants to select the whole folder again. Is this the case?

kspearrin:
Yea, we don't want to hide the select all option. The only time that would be acceptable is if all items were already selected.

ysshah:
Makes sense. I'll close this pull request then. Thanks!

PR #119, opened 2017-10-30
Focus by default the vault search input.

bfabio:


kspearrin:
Thanks for the contribution.

PR #120, opened 2017-10-30
add robots.txt disallowing indexing by crawlers

chipitsine:
if someone installs personal vault, it is not something he wants
to be indexed by crawlers

### DIFF
--- a/src/robots.txt
+++ b/src/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /


kspearrin:
I can understand this for self-hosted installation, but we do not want to stop a search engine from crawling `vault.bitwarden.com`. The gulp dist tasks  will probably need to be updated for that.

Mart124:
Some news regarding this ?
Many thanks 👍

ffflorian:
@chipitsine

You could also add the following to your `bwdata/nginx/default.conf` file:

```nginx
location = /robots.txt {
  add_header Content-Type text/plain;
  return 200 "User-agent: *\nDisallow: /\n";
}
```

After that, just run `bitwarden.sh restart` and the `robots.txt` file is accessible on your self-hosted installation.

setyb:
@kspearrin Wrote:

> I can understand this for self-hosted installation, but we do not want to stop a search engine from crawling `vault.bitwarden.com`.

Why is this?  So people can search for the vault using a search engine?  That's obviously a bad practice, but maybe something people actually do?

nextlineIO:
> @chipitsine
> 
> You could also add the following to your `bwdata/nginx/default.conf` file:
> 
> ```nginx
> location = /robots.txt {
>   add_header Content-Type text/plain;
>   return 200 "User-agent: *\nDisallow: /\n";
> }
> ```
> 
> After that, just run `bitwarden.sh restart` and the `robots.txt` file is accessible on your self-hosted installation.

Doesn't the ../bwdata/nginx/default.conf reset after update?  Is this something you have to add to this file every time you update?

nicolaevladescu:
Hi.

Maybe somebody else would find this helpful.

I worked around this issue via CloudFlare (free plan) with a simple Worker that overrides only `mydomain.com/robots.txt` route, the only caveat is that you need to enable cloud protection for your `A` or `CNAME` record and obviously host your DNS on CloudFlare.

Worker code:

```
addEventListener('fetch', event => {
  event.respondWith(handleRequest(event.request))
})

/**
 * Respond to the request
 * @param {Request} request
 */
async function handleRequest(request) {
  return new Response('User-agent: *\nDisallow: /\n', {status: 200})
}
```

P.S. I also added a firewall rule that block known bots, just in case.

eliykat:
Hi @chipitsine, thanks for your submission! If you're still interested in progressing this PR, please:
* sign the Contributor License Agreement (see the link provided by the CLAassistant bot above)
* review/respond to any outstanding change requests or questions
* reply here and tag me (@) so that I can mark it for further review

As this PR is more than 12 months old, I'll close it for now. However, this is just housekeeping. We encourage our community members to get involved and make contributions, so you can comment here even after it's closed and I'll reopen it if required.

PR #144, opened 2017-12-04
vaultAddCipherController.js: secureNote Type is int not string

philhug:
SecureNote Type is defined as int not as string, send int in JSON

### DIFF
--- a/src/app/vault/vaultAddCipherController.js
+++ b/src/app/vault/vaultAddCipherController.js
@@ -15,7 +15,7 @@
             identity: {},
             card: {},
             secureNote: {
-                type: '0'
+                type: 0
             }
         };
 


kspearrin:
Seems right to me, but curious how you even came across this?

philhug:
@kspearrin found this while working on a golang server. go is less tolerant at parsing JSON. (https://github.com/philhug/bitwarden-go)
btw: are you familiar with openapi/swagger (https://www.openapis.org/) would you be interested in having the API documented/specified?

kspearrin:
There are ways to implement swagger pretty easily in .NET Core project with  https://github.com/domaindrivendev/Swashbuckle . We can look into it.

PR #145, opened 2017-12-04
Changed MAC comparison to constant time comparison

Traur:
I changed the algorithm for mac comparison to be in constant time. That means comparing every single byte of each mac with one another and not failing after one miss. 

If you exit after the first miss the algorithm takes longer if there are more matching bytes and this may be an issue for timing attacks. Checking every byte thus makes this constant time.

### DIFF
--- a/src/app/services/cryptoService.js
+++ b/src/app/services/cryptoService.js
@@ -779,14 +779,14 @@ angular
 
                     var arr1 = new Uint8Array(mac1);
                     var arr2 = new Uint8Array(mac2);
-
+                    var macsEqualConstantTime = true
                     for (var i = 0; i < arr2.length; i++) {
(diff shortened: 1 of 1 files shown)


kspearrin:
We already do a double HMAC verification here, which covers the timing attack vector.  https://www.nccgroup.trust/us/about-us/newsroom-and-events/blog/2011/february/double-hmac-verification/

Traur:
I can see you are doing this in the macsEqual() function above, but not in the macsEqualWC(). I changed the latter one.

kspearrin:
It is done in this function as well.

Traur:
Oooh now I get it. You're calculating the HMAC twice and then compare the result byte after byte. That's actually a clever idea. Nevermind then.

PR #177, opened 2018-02-16
Full width layout

nblackburn:
Seems a shame to constrain your layout to 1250px when you can have a layout that can take full advantage of the space available.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -61,7 +61,7 @@
     <link rel="mask-icon" href="images/icons/safari-pinned-tab.svg" color="#3c8dbc">
     <link rel="manifest" href="manifest.json">
 </head>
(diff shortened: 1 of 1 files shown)


kspearrin:
In what situations is the constrained layout reducing visibility of content? We've implemented most things to ensure that isn't an issue. Stretching the layout to non-predictable widths reduces the ability for us to accurately position things in certain cases.

It seems like you're just trading one set of whitespace for another.

nblackburn:
@kspearrin I don't believe i said it was reducing visibility but can't think of a use case where you would use anything beyond the scope of the template you are using that would require you to add something custom for that to even be a concern.

You are right about the trade of whitespace but at least with the full width layout, it gives the impression that it belongs to the application rather than it being outside of it.

There are cases for contained layouts but being as this template was designed to work at any resolution, i feel it works best to play to it's strengths.

kspearrin:
Web vault 2.0 has a different layout now.

PR #211, opened 2018-06-16
Mask Card Security Code . (bitwarden/desktop issue 78)

neilsb:
By default shows Card Security Code masked in the same way as Passwords. (bitwarden/desktop issue 78)


kspearrin:
I am going to leave this PR open since the web vault is currently undergoing a complete re-write in the `angular` branch which makes these changes irrelevant.

kspearrin:
If you want to submit a PR for the angular branch I can accept that. It should be close to the desktop and browser implementations since we now use jslib there.

PR #213, opened 2018-06-16
Recognise Cards & Secure Note in Enpass Import

neilsb:
Try to identify Secure Notes and Card Details from Enpass export and assign the correct types in BitWarden.

Since importing is not in the Angular branch yet, adding to master


kspearrin:
This is now implemented in the new importer here: https://github.com/bitwarden/jslib/commit/b4846e5feaf0d5a6122950e0b432d792a443015f

PR #233, opened 2018-07-31
Removed requirement to load JavaScript from js.braintreegateway.com

lightswitch05:
I removed the requirement to load JavaScript from js.braintreegateway.com by adding that javascript file to be hosted within your own domain. The logic to only load run JavaScript file as needed remains. I believe this improves security and stability since the file is now hosted within your own domain and modifications to it should only occur during upgrades.

package-lock.json had a lot more changes then I expected from this update. The change goes from exact dependency versions in to ranges instead. I'm not sure what caused this, I have NPM version 6.2.0


kspearrin:
Can you revert the lock file changes please? I'll regenerate from my end.

lightswitch05:
@kspearrin done

kspearrin:
Any reason we're using this package over https://www.npmjs.com/package/braintree-web-drop-in ? Is ./node_modules/braintree-web/dist/braintree.js the actual dropin file?

lightswitch05:
@kspearrin actually it looks like drop-in does do more for you then the standard project. Bad news that [braintree/braintree-web-drop-in](https://github.com/braintree/braintree-web-drop-in) purposefully doesn't include an already built js file: https://github.com/braintree/braintree-web-drop-in/issues/233 to save 186kb

lightswitch05:
Neither 'https://js.stripe.com/v2/' or 'https://js.braintreegateway.com/web/dropin/1.4.0/js/dropin.min.js' can be used with Subresource Integrity (SRI) since they don't have CORS enabled (required for SRI). With braintree disagreeing that a pre-compiled version would be useful, and strip not providing an external package at all (as far as I can tell), what do you think about having a build step to download these scripts so they can be hosted internally?

kspearrin:
Yes, I am aware of that, which is why they don't have SRI hashes today. It might be possible for Braintree since they are versioned, but won't be possible with Stripe since they manage versioning internally and you never know when the script gets updated. I've already talked to their dev team. This is how they require the script to be used.

lightswitch05:
This will be possible with a [new release](https://github.com/braintree/braintree-web-drop-in/issues/233#issuecomment-412088501) from braintree. Release timeline:

> Probably next week.

PR #259, opened 2018-09-11
Removed requirement to load JavaScript from js.braintreegateway.com

lightswitch05:
Braintree was kind enough to supply a built version of their drop-in library in their node package via https://github.com/braintree/braintree-web-drop-in/issues/233 

By pulling this library in through a static build process, it removes the run-time dependency braintreegateway.com also reducing the attack surface of the vault


package.json
```diff
@@ -32,6 +32,7 @@
     "@types/papaparse": "4.1.33",
     "@types/webcrypto": "^0.0.28",
     "angular2-template-loader": "^0.6.2",
+    "braintree-web-drop-in": "^1.13.0",
```
kspearrin: Shouldn't this be a regular dependency and not a dev?

lightswitch05: I guess? The JS gets copied over during build time. I'm happy to update it if you want

kspearrin: The `qrious` library you are modeling this after is in dependencies. please add it there too.

kspearrin:
I notice that the dropin is now unminified and 600kb in size. That's pretty large vs before. Any way we can get a min version?

lightswitch05:
Seems reasonable, I asked here https://github.com/braintree/braintree-web-drop-in/issues/233#issuecomment-420439567

lightswitch05:
After this change goes live, I believe the vault's CSP header could be updated:

Current: 
````
content-security-policy: default-src 'self'; script-src 'self' 'sha256-ryoU+5+IUZTuUyTElqkrQGBJXr1brEv6r2CA62WUw8w=' https://js.stripe.com https://js.braintreegateway.com https://www.paypalobjects.com; style-src 'self' 'unsafe-inline' https://assets.braintreegateway.com https://*.paypal.com; img-src 'self' data: https://icons.bitwarden.net https://*.paypal.com https://www.paypalobjects.com https://q.stripe.com https://haveibeenpwned.com https://www.gravatar.com; child-src 'self' https://js.stripe.com https://assets.braintreegateway.com https://*.paypal.com https://*.duosecurity.com; frame-src 'self' https://js.stripe.com https://assets.braintreegateway.com https://*.paypal.com https://*.duosecurity.com; connect-src 'self' wss://notifications.bitwarden.com https://notifications.bitwarden.com https://cdn.bitwarden.com https://haveibeenpwned.com https://api.pwnedpasswords.com https://api.stripe.com https://www.paypal.com https://api.braintreegateway.com https://client-analytics.braintreegateway.com https://*.braintree-api.com https://www.google-analytics.com;
````

* I believe https://js.braintreegateway.com can be removed from `script-src`. I'm not sure about the other braintree endpoints. I never see `assets.braintreegateway.com` or `client-analytics.braintreegateway.com` - but I also have not purchased anything

lightswitch05:
That discussion with braintree about injecting iframes and scripts brought up some good points. I wonder how hard it would be to move all payment processing to a standalone page that is outside of the vault. Moving it out not reduces the attack surface of the vault, but would also allow for a simpler and more robust CSP header

kspearrin:
There are future plans to create a "billing portal" that is separate from the web vault, however, for convenience reasons there will likely always be the ability to make purchases from directly within the web vault as well.

PR #276, opened 2018-10-14
Accessibility Refactor of Checkboxes on Collection & Share Modals

acespacecase:
I noticed a couple modals that have checkboxes that had odd behavior: you could click on the background of the row or the checkbox, but not the text itself. While looking at the code, I also noticed there was a way to restructure the HTML to be more accessible for users by using fieldset, legend, and arias. By doing this, I also was able to refactor the click behavior to use HTML5's checking and unchecking behavior (label + input), rather than a Typescript function (check(c) function).


kspearrin:
Hi. Thanks for the PR. At the moment, clicking the name is not checking the box because `appStopProp` is on the wrapping `span`. Simply removing this `span` would resolve that issue.

We want the table with row-click checking to remain in place since that is a standard used in several areas of the web vault. To change that would require us to make any changes in all places where we currently use `table-list-checkbox` to be consistent. I am not sure that is something we want to change at this time.

I am happy to accept a PR that removes the `appStopProp` in this case though.

acespacecase:
Sounds good! Thanks for the reply.

PR #279, opened 2018-10-18
Add .gitattributes file

SoulSeekkor:
Added .gitattributes file to files requiring LF endings are properly checked out on Windows.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+*.sh eol=lf
+.dockerignore eol=lf
+dockerfile eol=lf


kspearrin:
Could we just make everything EOL LF?

SoulSeekkor:
That would work when developing on non-Windows (I'm always on Windows at work and home), with Git installed on Windows with the defaults of checkout Windows-style/commit Unix-style this breaks anything docker or bash script related for me, if I do a pull I have to add this file commit then blow away and re-pull in order to fix it or manually edit every affected file.

This isn't quite as bad with this repo but it's a huge headache on the core repo for sure, this is just a fail-safe to ensure the files that need to be LF are always checked out LF (even when the rest is fine as CRLF on Windows).  Might save some folk the headache I ran into when initially building docker images failed the first time due to the incorrect line-endings.

PR #288, opened 2018-11-11
Proposal: Add field for birthday to identity cipher

SoulSeekkor:
I'd like to propose adding the "Birthday" field to the identity cipher.  I've done the work for the jslib, core, and web repos and can do the pull request for those if this would be accepted.  The only thing I wasn't sure on was the other language translations.

### DIFF
--- a/src/app/vault/add-edit.component.html
+++ b/src/app/vault/add-edit.component.html
@@ -214,6 +214,10 @@
                                 <option *ngFor="let o of identityTitleOptions" [ngValue]="o.value">{{o.name}}</option>
                             </select>
                         </div>
+                        <div class="col-4 form-group">
+                            <label for="idBirthday">{{'birthday' | i18n}}</label>
+                            <input id="idBirthday" class="form-control" type="date" name="Identity.Birthday" [(ngModel)]="cipher.identity.birthday">
+                        </div>
                     </div>
                     <div class="row">
                         <div class="col-4 form-group">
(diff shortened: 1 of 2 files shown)


SoulSeekkor:
Hmm, I see where you are going there, I had tested the loading and saving but only in Chrome. I'll have to do some digging and testing to see if that would cause any issue loading the value in other pieces like the extension and desktop app, you are worried that other applications may load the source data and format it or display it incorrectly?

I suppose the easiest fix would be to have the input type changed if nothing else, though the date input is very nice to use.

kspearrin:
Well, if you plan to use this date field in any meaningful way (such as autofilling it) it would have to be formatted in a way that could reliably be parsed and used for day, month , and year.

SoulSeekkor:
I understand, I hadn't even really been considering from an auto fill standpoint to be honest, it was purely for saving what felt like a missing piece of information for an identity.

Knowing to look at it from an autofill perspective however, I'll take a second look to see if it's a viable field to have added.

As a side note, I had a couple ideas for actual new features, what is the preferred method to propose those? Chat, email, or perhaps a github issue?

kspearrin:
Dev chatroom or GitHub issues would be fine.

eliykat:
Hi @SoulSeekkor, thanks for your submission! If you're still interested in progressing this PR, please:
* review/respond to any outstanding change requests or questions
* reply here and tag me (@) so that I can mark it for further review

As this PR is more than 12 months old, I'll close it for now. However, this is just housekeeping. We encourage our community members to get involved and make contributions, so you can comment here even after it's closed and I'll reopen it if required.

PR #289, opened 2018-11-11
Proposal: Add "Value 2" to custom fields

SoulSeekkor:
I'd like to propose adding a "Value 2" to the custom fields.  The easiest case for this is security questions, when trying to add those to logins to save them (and be able to use random ones...perhaps we could add a randomize button next to the two values, though that would use a lot of space) there isn't a way to do so that makes sense.  You can name it security question but you can't save the question/answer pair then, you can set the name as the question but that doesn't make sense either.  This would allow cases where you need to save a name/value pair as a custom field but still be able to give it a proper name.

I've done the work for the jslib, core, and web repos and can do the pull request for those if this would be accepted.  The only thing I wasn't sure on was the other language translations and what may need to get changed for the if statement in jslib cipher.service.ts on line 103 "if (existingCipher.hasFields)".

### DIFF
--- a/src/app/vault/add-edit.component.html
+++ b/src/app/vault/add-edit.component.html
@@ -309,7 +309,7 @@
                 <h3 class="mt-4">{{'customFields' | i18n}}</h3>
                 <ng-container *ngIf="cipher.hasFields">
                     <div class="row" appBoxRow *ngFor="let f of cipher.fields; let i = index">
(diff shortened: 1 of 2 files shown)


SoulSeekkor:
Hmm, fair enough. I think prefixing the name with security question is a fair way of working around that.

PR #342, opened 2019-02-03
Fix disappearing fields

ShirokaiLon:
This fixes https://github.com/bitwarden/browser/issues/847 by adding the `trackBy` option to the `ngFor` directives used for URIs and custom fields.

https://github.com/bitwarden/jslib/pull/27 is required.

### DIFF
--- a/src/app/vault/add-edit.component.html
+++ b/src/app/vault/add-edit.component.html
@@ -105,7 +105,7 @@
                         </div>
                     </div>
                     <ng-container *ngIf="cipher.login.hasUris">
-                        <div class="row" appBoxRow *ngFor="let u of cipher.login.uris; let i = index">
+                        <div class="row" appBoxRow *ngFor="let u of cipher.login.uris; let i = index; trackBy:trackByFunction">
                             <div class="col-7 form-group">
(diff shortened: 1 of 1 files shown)


src/app/vault/add-edit.component.html
```diff
@@ -308,7 +308,7 @@ <h2 class="modal-title">{{title}}</h2>
                 </div>
                 <h3 class="mt-4">{{'customFields' | i18n}}</h3>
                 <ng-container *ngIf="cipher.hasFields">
-                    <div class="row" appBoxRow *ngFor="let f of cipher.fields; let i = index">
+                    <div class="row" appBoxRow *ngFor="let f of cipher.fields; let i = index; trackBy:trackByFunction">
```
kspearrin: You're dealing with different types here. So you should probably change the function in jslib to `any` type.

ShirokaiLon: You're right, of course. It was only later on that I remembered that custom fields were an issue, too :roll_eyes:

kspearrin: No worries. I fixed it :)

PR #384, opened 2019-04-25
Add a "launch site" button directly to the list of ciphers

TastesLikeTurkey:


TastesLikeTurkey:
Currently launching a site takes at least 3 clicks; one to open the card, one to launch the site and another to close the card. This enables launching the site with just a single click.

kspearrin:
I am not really a fan of adding another button to the row there. What if we took the same approach as the browser extension and added a double click to launch the website?

TastesLikeTurkey:
@kspearrin My only gripe with the double-click in the extension is that when you are actually just single-clicking on the item it tends to feel a bit sluggish because of the delay while it waits for a second click. But I guess it's probably not that big of a deal... I'll try replicating that here.

TastesLikeTurkey:
@kspearrin Would you be opposed to adding a "Launch" option to the dropdown menu? IMO that would still make it much more convenient to launch a site while avoiding the aforementioned sluggishness.

kspearrin:
Sure, that would be better.

src/app/vault/ciphers.component.html
```diff
@@ -34,6 +34,11 @@
                                 <i class="fa fa-fw fa-clipboard"></i>
                                 {{'copyPassword' | i18n}}
                             </a>
+                            <a class="dropdown-item" href="#" appStopClick
+                                *ngIf="c.type === cipherType.Login && c.login.canLaunch" (click)="launch(c.login)">
```
kspearrin: You appear to be passing a `LoginView` object here, yet your `lunch` function is accepting a `LoginUriView`.

TastesLikeTurkey: Oops! Should be fixed now. Sorry, this is my first foray into TypeScript and Angular...

PR #445, opened 2020-01-25
Stop showing score 3 passwords as weak passwords

naokiri:
* Summary
Stops showing score 3 passwords in Weak Passwords Report page.
https://vault.bitwarden.com/#/tools/weak-passwords-report

* Reason
In zxcvbn library that scores the password, score 3 is considered "safely unguessable" (c.f. https://github.com/dropbox/zxcvbn#usage )
Showing too many passwords in the list distracts users and prevents from focusing on really weak passwords.


Mart124:
Mmmmh I'm rather happy that Web Vault shows score 3 passwords...
Perhaps we will be able to use the password strength policy score as the threshold for weak passwords report @kspearrin ?

PR #446, opened 2020-01-25
Sort weak passwords by severity

naokiri:
* Summary
In the weak passwords report screen, sort the passwords by weakness to show 'Danger' level passwords on top of the list.
https://vault.bitwarden.com/#/tools/weak-passwords-report

* Reason
Idea based on this feature request
https://community.bitwarden.com/t/weak-passwords-sorting/10005

I think this is one way to emphasize weaker passwords and encourage to change passwords to a better one.


dmcfaddengalway (dismissed):
Good job!

src/app/tools/weak-passwords-report.component.ts
```diff
@@ -20,6 +20,15 @@ import { CipherReportComponent } from './cipher-report.component';
     templateUrl: 'weak-passwords-report.component.html',
 })
 export class WeakPasswordsReportComponent extends CipherReportComponent implements OnInit {
+
+    private static hasUsername(c: CipherView): boolean {
```
cscharf, line 24: Any reason these need to be static? I would think you could have made them local `const` functions within the `passwordStrengthCache()` method, OR at least instance local (not static).

cscharf (commented):
Thanks for the PR! These looks pretty good, just some minor feedback.

src/app/tools/weak-passwords-report.component.ts
naokiri: Instance local doesn't make sense because they don't touch anything instance local, and `passwordStrengthCache` is a `Map` specific to this class so I wanted those utility functions tied to this class.

Making them local const function within the `setCiphers()` method is acceptable, but we have to be careful because they are called in the closure of `allCiphers.forEach`. Probably I have to change either variable or function name of `hasUsername` not to capture the name.

Among those options I think making them static functions are better.

naokiri: > fitting greater body of code

In that sense I'd choose to make them local const function inside the `setCiphers()` method. I don't have time and environment to touch here right now so I'll push the fix of this branch next week.

cscharf (approved):
Looks good, thank you!

PR #457, opened 2020-02-10
Implement Clone item functionality (personal/org)

vincentsalucci:
> Add the ability to clone items on the web vault.  
**Personal Vault Context**
Personal items can be cloned 
Allow ownership selection (collections that are "read only" will not be visible)
Restrict organizational item cloning 
**Organizational Vault Context**
Organizational items can be cloned
Allow collections to be chosen during creation (collections that are "read only" will not be visible)

- **Organization: add-edit.component.ts**: Overridden saveCipher method will properly route add/edit call based on admin status
- **Organization: vault.component.html**: Added databinding for cloneCipher callback
- **Organization: vault.component.ts**:  Added cloneCipher method. Sets the necessary add/edit variables for proper adjustment
- **add-edit.component.html**: Allowed ownership/collection UI visibility and removed the delete button from cloning operations
- **ciphers.component.html**: Added clone option to menu
- **ciphers.component.ts**: Added clone callback from databinding
- **vault.component.html**: Added databinding for cloneCipher callback
- **vault.component.ts**: Added clone callback from databinding
- **messages.json**: Added "Clone" string

**Personal Vault**
0-myvault-clone
1-myvault-clone-item-default
2-myvault-clone-item-to-org
3-myvault-clone-success
4-myvault-clone-from-org-denied

**Organization Vault**
5-orgvault-clone-admin
6-orgvault-clone-item-admin
7-orgvault-clone-success-admin
8-orgvault-clone-collections-user

### DIFF
--- a/src/app/organizations/vault/vault.component.ts
+++ b/src/app/organizations/vault/vault.component.ts
@@ -263,6 +263,18 @@ export class VaultComponent implements OnInit, OnDestroy {
         return childComponent;
     }
 
+    cloneCipher(cipher: CipherView) {
+        const component = this.editCipher(cipher);
+        component.cloneMode = true;
+        component.organizationId = this.organization.id;
+        if (this.organization.isAdmin) {
+            component.collections = this.groupingsComponent.collections.filter((c) => !c.readOnly);
+        }
+        // Regardless of Admin state, the collection Ids need to passed manually as they are not assigned value
+        // in the add-edit componenet
+        component.collectionIds = cipher.collectionIds;
+    }
+
     async viewEvents(cipher: CipherView) {
         if (this.modal != null) {
             this.modal.close();
--- a/src/app/vault/ciphers.component.html
+++ b/src/app/vault/ciphers.component.html
@@ -52,6 +52,11 @@
                                 <i class="fa fa-fw fa-paperclip" aria-hidden="true"></i>
                                 {{'attachments' | i18n}}
                             </a>
+                            <a class="dropdown-item" href="#" appStopClick
+                                *ngIf="(!organization && !c.organizationId) || organization" (click)="clone(c)">
+                                <i class="fa fa-fw fa-clone" aria-hidden="true"></i>
+                                {{'clone' | i18n}}
+                            </a>
                             <a class="dropdown-item" href="#" appStopClick *ngIf="!organization && !c.organizationId"
                                 (click)="share(c)">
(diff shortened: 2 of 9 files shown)


src/app/organizations/vault/vault.component.ts
```diff
@@ -263,6 +263,18 @@ export class VaultComponent implements OnInit, OnDestroy {
...
+    cloneCipher(cipher: CipherView) {
+        const component = this.editCipher(cipher);
+        component.cloneMode = true;
+        component.organizationId = this.organization.id;
+        if (this.organization.isAdmin) {
+            component.collections = this.groupingsComponent.collections.filter((c) => !c.readOnly);
+        }
+        // Regardless of Admin state, the collection Ids need to passed manually as they are not assigned value
+        // in the add-edit componenet
+        component.collectionIds = cipher.collectionIds;
```
kspearrin: What if the cloning user doesn't have access to assign the items to the existing `cipher.collectionIds`? I say we just use the same logic as the addCipher:

```
if (this.collectionId != null) {
	component.collectionIds = [this.collectionId];
}
```

Maybe I am missing something though.

src/app/vault/ciphers.component.html
```diff
@@ -88,4 +93,4 @@
                 <i class="fa fa-plus fa-fw"></i>{{'addItem' | i18n}}</button>
         </ng-container>
     </div>
-</ng-container>
+</ng-container>
```
kspearrin: Why does this happen? I noticed it on another file too. Removing the empty new line at EOF?

src/locales/en/messages.json
```diff
@@ -2956,5 +2956,8 @@
   },
   "minLength": {
     "message": "Minimum Length"
+  },
+  "clone": {
+    "message": "Clone"
   }
-}
+}
```
kspearrin: again here

src/app/organizations/vault/vault.component.ts
vincentsalucci: In the add flow as a nicety, we assume that if they are viewing a specific collection and hit "add item" button, they want to add it to the currently view collection. 

However, it is different during the clone flow because technically we don't care about what's being viewed, rather, we care about the collections specific to the cipher being cloned.

To address your question: because we made the change to allow clone mode to manually filter the passed collections, if a user does not have access to assign items to any collections, they will be filtered out from the list.

PR #472, opened 2020-02-28
Show policy in effect banner for password generator

vincentsalucci:
## Code Changes
- **password-generator.component.html**: Added `app-callout` banner and made visible when a policy is in effect
- **messages.json**: Added `passwordGeneratorPolicyInEffect` string

## Screenshots
0-pg-policy-banner

### DIFF
--- a/src/app/tools/password-generator.component.html
+++ b/src/app/tools/password-generator.component.html
@@ -1,6 +1,9 @@
 <div class="page-header">
     <h1>{{'passwordGenerator' | i18n}}</h1>
 </div>
+<app-callout type="primary" *ngIf="hasPolicyInEffect()">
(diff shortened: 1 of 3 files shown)


src/app/tools/password-generator.component.html
```diff
@@ -1,6 +1,9 @@
 <div class="page-header">
     <h1>{{'passwordGenerator' | i18n}}</h1>
 </div>
+<app-callout type="primary" *ngIf="hasPolicyInEffect()">
```
kspearrin: Let's change it to `type="info"`

vincentsalucci: Updated

PR #506, opened 2020-03-30
[Auto-Logout] Implement upstream changes

vincentsalucci:
## Code Changes
- **jslib**: Update from 31a2574 -> 28e3fff
- **lock.component.ts**: Updated import. Updated constructor. Updated super call.
- **app.component.ts**: Updated import. Updated constructor. Updated necessary calls to the new `vaultTimeoutService`
- **services.module.ts**: Updated imports. Updated instantiation for new name and added log out callback action. Updated dependencies to use new abstraction/service.
- **unauth-guard.service.ts**: Updated import. Updated constructor. Updated necessary calls to the new `vaultTimeoutService`
- **options.component.html**: Updated name dependencies. Added `vaultTimeoutAction` options.
- **options.component.ts**: Updated import. Updated variable names to reflect `vaultTimeout..` pattern. Updated constructor. Updated necessary calls to the new `vaultTimeoutService`. Saving new selected action.
- **messages.json**: Updated old strings. Added new strings.
- **htmlStorage.service.ts**: Updated `localStorageKeys` to include both of the vault timeout options. Side note: the original `lockOptionKey` was accidentally listed twice. Updated `init` method defaults.

## Screenshots
0-auto-logout-options
1-auto-logout-action

### DIFF
--- a/src/app/services/services.module.ts
+++ b/src/app/services/services.module.ts
@@ -38,7 +38,6 @@ import { EventService as EventLoggingService } from 'jslib/services/event.servic
 import { ExportService } from 'jslib/services/export.service';
 import { FolderService } from 'jslib/services/folder.service';
 import { ImportService } from 'jslib/services/import.service';
-import { LockService } from 'jslib/services/lock.service';
 import { NotificationsService } from 'jslib/services/notifications.service';
 import { PasswordGenerationService } from 'jslib/services/passwordGeneration.service';
 import { PolicyService } from 'jslib/services/policy.service';
@@ -49,6 +48,7 @@ import { SyncService } from 'jslib/services/sync.service';
 import { TokenService } from 'jslib/services/token.service';
 import { TotpService } from 'jslib/services/totp.service';
 import { UserService } from 'jslib/services/user.service';
+import { VaultTimeoutService } from 'jslib/services/vaultTimeout.service';
 import { WebCryptoFunctionService } from 'jslib/services/webCryptoFunction.service';
 
 import { ApiService as ApiServiceAbstraction } from 'jslib/abstractions/api.service';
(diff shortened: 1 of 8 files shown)


src/locales/en/messages.json
```diff
@@ -3040,5 +3040,17 @@
   },
   "userPreference": {
     "message": "User Preference"
+  },
+  "vaultTimeoutAction": {
```
vincentsalucci: @kspearrin @cscharf @mportune-bw Any verbiage suggestions? I followed existing patterns but I'm always open for changes that everyone agrees on.

cscharf: The only thing I would say is that the other properties, variables, methods, etc. exclude the term "vault" as they already apply to the vault in principal. Since this replaces 1st class citizen concepts like "lock" which were only "lock" or "lockOptions", etc. perhaps "timeoutAction" would be enough, vs. the more verbose/explicit naming convention implemented here. Just my $0.03 (adjusted for inflation).

kspearrin: I agree we're a little verbose on the naming here, but that's outside the scope of this PR. We can consider renaming things "timeoutService", "timeout", "timeoutAction", etc in a future PR if it makes sense to everyone.

src/app/services/services.module.ts
```diff
@@ -108,8 +108,9 @@ const folderService = new FolderService(cryptoService, userService, apiService,
 const collectionService = new CollectionService(cryptoService, userService, storageService, i18nService);
 searchService = new SearchService(cipherService, platformUtilsService);
 const policyService = new PolicyService(userService, storageService);
-const lockService = new LockService(cipherService, folderService, collectionService,
-    cryptoService, platformUtilsService, storageService, messagingService, searchService, userService, null);
+const vaultTimeoutService = new VaultTimeoutService(cipherService, folderService, collectionService,
+    cryptoService, platformUtilsService, storageService, messagingService, searchService, userService, null,
+    async () => messagingService.send('logout', { expired: true }));
```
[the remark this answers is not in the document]
vincentsalucci: That makes sense. While implementing, I convinced myself that arguments could be made for both sides of the conversation. Since the vault timeout could be interpreted as a user's access being 'expired'.  Regardless, will update to use false.

Mart124:
@vincentsalucci @kspearrin while reading this PR, and idea, add the lock / logout options to the organization policies, so that security can be enforced on this ?
Perhaps you already have this in mind :)
Thx 👍

PR #509, opened 2020-04-01
Fixed Windows Symlink command in package.json for npm scripts

cscharf:
Tried setting up the symlink on Windows and noticed the npm script failed with unknown command... fixed the script to get that working.

rmdir vs. rm, /S /Q vs. -rf.

symlink pulled latest from jslib /master.


cscharf:
Abandoning PR

PR #511, opened 2020-04-08
[Soft Delete] - Added trash and related functionality to web vault

cscharf:
## Objective
Incorporate the new APIs for soft deleting and restoring cipher items into the web vault client. This includes a UX that presents a new logical "Trash" bin under "Favorites" on _My Vault_ view and under "All Items" in the _Organization_ view.

### Design Theory (UX)
Easy to local and use **Trash**. When deleting an item anywhere in the interface (from the vault list, bulk action or in the edit/detail view) this item will now be "soft deleted" (essentially moved to the Trash). When performing searches, etc. the scope of that search will always either be "Trash" or anything else (when selected). Context will not cross between items in Trash from the rest of the vault, including "All Items".

image
* Context under Trash changes to show delete options as "Permanent" vs. just Delete.
* A new option to "Restore" selected items appears under bulk and individual options
image
* Item detail/open item is now a "read-only" view of that item
* When viewing detail of a deleted item (in trash) will change the delete function to be "permanent"
* The "Save" button is replaced with the only action available other than cancel/permanently delete as "Restore"
image
* Organization view has similar context for trash
image

### DIFF
--- a/src/app/vault/add-edit.component.html
+++ b/src/app/vault/add-edit.component.html
@@ -12,7 +12,8 @@
                 <div class="row" *ngIf="!editMode">
                     <div class="col-6 form-group">
                         <label for="type">{{'whatTypeOfItem' | i18n}}</label>
-                        <select id="type" name="Type" [(ngModel)]="cipher.type" class="form-control">
+                        <select id="type" name="Type" [(ngModel)]="cipher.type" class="form-control" 
+                            [disabled]="cipher.isDeleted">
                             <option *ngFor="let o of typeOptions" [ngValue]="o.value">{{o.name}}</option>
                         </select>
                     </div>
@@ -21,11 +22,12 @@
                     <div class="col-6 form-group">
                         <label for="name">{{'name' | i18n}}</label>
                         <input id="name" class="form-control" type="text" name="Name" [(ngModel)]="cipher.name"
-                            required>
+                            required [disabled]="cipher.isDeleted">
                     </div>
                     <div class="col-6 form-group" *ngIf="!organization">
                         <label for="folder">{{'folder' | i18n}}</label>
-                        <select id="folder" name="FolderId" [(ngModel)]="cipher.folderId" class="form-control">
+                        <select id="folder" name="FolderId" [(ngModel)]="cipher.folderId" class="form-control" 
+                            [disabled]="cipher.isDeleted">
                             <option *ngFor="let f of folders" [ngValue]="f.id">{{f.name}}</option>
                         </select>
                     </div>
@@ -37,8 +39,8 @@
                             <label for="loginUsername">{{'username' | i18n}}</label>
                             <div class="input-group">
                                 <input id="loginUsername" class="form-control" type="text" name="Login.Username"
-                                    [(ngModel)]="cipher.login.username" appInputVerbatim>
-                                <div class="input-group-append">
+                                    [(ngModel)]="cipher.login.username" appInputVerbatim [disabled]="cipher.isDeleted">
+                                <div class="input-group-append" *ngIf="!cipher.isDeleted">
                                     <button type="button" class="btn btn-outline-secondary"
                                         appA11yTitle="{{'copyUsername' | i18n}}"
                                         (click)="copy(cipher.login.username, 'username', 'Username')" tabindex="-1">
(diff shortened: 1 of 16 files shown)


src/app/vault/add-edit.component.html
```diff
@@ -85,7 +88,7 @@ <h2 class="modal-title" id="cipherAddEditTitle">{{title}}</h2>
                             </div>
                         </div>
                     </div>
-                    <div class="row">
+                    <div class="row" *ngIf="!cipher.isDeleted">
```
kspearrin, line 91: Why do we disable some things, but here you hide it?

src/app/vault/add-edit.component.html
```diff
@@ -132,9 +135,9 @@ <h2 class="modal-title" id="cipherAddEditTitle">{{title}}</h2>
                                 <label for="loginUri{{i}}">{{'uriPosition' | i18n : (i + 1)}}</label>
                                 <div class="input-group">
                                     <input class="form-control" id="loginUri{{i}}" type="text"
-                                        name="Login.Uris[{{i}}].Uri" [(ngModel)]="u.uri"
+                                        name="Login.Uris[{{i}}].Uri" [(ngModel)]="u.uri" [disabled]="cipher.isDeleted"
                                         placeholder="{{'ex' | i18n}} https://google.com" appInputVerbatim>
-                                    <div class="input-group-append">
+                                    <div class="input-group-append" *ngIf="!cipher.isDeleted">
```
kspearrin, line 140: I don't see any harm in leaving these available still.

src/app/vault/add-edit.component.html
```diff
@@ -492,19 +500,20 @@ <h3 class="mt-4">{{'collections' | i18n}}</h3>
             <div class="modal-footer">
                 <button type="submit" class="btn btn-primary btn-submit" [disabled]="form.loading">
                     <i class="fa fa-spinner fa-spin" title="{{'loading' | i18n}}" aria-hidden="true"></i>
-                    <span>{{'save' | i18n}}</span>
+                    <span>{{(cipher.isDeleted ? 'restore' : 'save') | i18n}}</span>
```
kspearrin, line 503: I see why you added the restore check on submit now. I guess this works.

src/locales/en/messages.json
```diff
@@ -3053,5 +3053,87 @@
...
+    "message": "Restored Items"
+  },
+  "restoreItemConfirmation": {
+    "message": "Are you sure you want to restore this item?"
+  },
+  "restoreItems": {
+    "message": "Restore items"
+  },
+  "restoreSelectedItemsDesc": {
+    "message": "You have selected $COUNT$ item(s) to restore. Are you sure you want to restore all of these items?",
```
vincentsalucci, line 3122: Is there a clean/built-in way to do quantity plurals? There was an issue opened up where someone complained that we don't use them and instead do things like the above `item(s)`. Just wondering if its something we could start doing where necessary.

src/app/vault/add-edit.component.html
cscharf: The general idea that we discussed was to make everything "read-only" for items in the trash, and leaving edit fields, add/remove buttons, etc. available on screen would give the impression that you can still do things with the cipher and have them persist (but they would not actually persist/save/update).

src/app/vault/add-edit.component.html
cscharf: Input fields, select lists, etc. would be disabled (read-only) where actions (add/remove) or rows with the entire purpose of adding new items like Urls or Custom fields, etc. are completely hidden/removed from view. We can discuss the general UX approach tomorrow so any adjustments can be made before we move on to other clients.

src/locales/en/messages.json
kspearrin: I have not figured that out yet. I know angular has some pluralize helpers but not sure how we can use that in the context of a long translated string like this.

kspearrin:
Let's review tomorrow @cscharf

src/app/vault/add-edit.component.html
cscharf: Fixed

src/app/vault/add-edit.component.html
cscharf: Fixed

cscharf:
@kspearrin , I've updated the edit view with the changes that would keep read-only behavior in-tact but disable editing when an item is deleted. Please re-review.

matt-h:
Thank you for this! This was one of the features that I was wanting in Bitwarden the most.

PR #524, opened 2020-05-01
fix: dependency issues

SrdjanMilic:
**What is fixed**

- @angular/compiler-cli@7.2.11 requires a peer of @angular/compiler@7.2.11...
- angular2-toaster@6.1.0 requires a peer of @angular/common@^6.0.0...
- angular2-toaster@6.1.0 requires a peer of @angular/compiler@^6.0.0...
- angular2-toaster@6.1.0 requires a peer of @angular/core@^6.0.0...
- bootstrap@4.3.1 requires a peer of popper.js@^1.14.7...

**Note:**
- `popper.js` package is deprecated, but it needs to stay as bootstrap still depends on it. 
- `@popperjs/core` is `popper.js` successor.

Some of the packages are upgraded during the process.


cscharf:
@SrdjanMilic , package updates in the TypeScript clients require consistency across each of the clients and jslib. For these updates can you please also update (where applicable) jslib, desktop, browser, (and the jslib submodule ref in cli)?
cc: @kspearrin

kspearrin:
We can likely hold off on this since the next major update we want to make will probably be to Angular 8.

SrdjanMilic:
So, I don't have to do nothing?

eliykat:
I assume this is no longer relevant as our dependencies have upgraded several versions since then and I haven't experienced any issues. Closing.

PR #525, opened 2020-05-02
fix: webpack.config.js plugin

SrdjanMilic:
**What is fixed**

Warning and error from npm and build log:
1. warning
`warning: Since webpack v4 the extract-text-webpack-plugin should not be used for css. Use mini-css-extract-plugin instead.`

2. error
`Entrypoint undefined = extract-text-webpack-plugin-output-filename...`

New plugin implementation documents:
[https://webpack.js.org/plugins/extract-text-webpack-plugin/#usage](https://webpack.js.org/plugins/extract-text-webpack-plugin/#usage)
[https://webpack.js.org/plugins/mini-css-extract-plugin/#advanced-configuration-example](https://webpack.js.org/plugins/mini-css-extract-plugin/#advanced-configuration-example)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4007,52 +4007,6 @@
         }
       }
     },
-    "extract-text-webpack-plugin": {
-      "version": "4.0.0-beta.0",
-      "resolved": "https://registry.npmjs.org/extract-text-webpack-plugin/-/extract-text-webpack-plugin-4.0.0-beta.0.tgz",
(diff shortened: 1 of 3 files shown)


kspearrin:
When we make changes like this, we need to apply it to all webpack projects to be consistent. browser, desktop, web.

SrdjanMilic:
Do you suggest that I should make changes and create PR for all those apps?

kspearrin:
Yes, please.

SrdjanMilic:
OK. I will do that.

PR #526, opened 2020-05-02
fix: icons, labels and check boxes vertical align

SrdjanMilic:
Check box elements, site icons and labels were not properly vertically aligned.  
Setting up the `line-height: 0;` on few elements, and also new properties for `.form-check-label` class, fix those issues.

### DIFF
--- a/src/scss/styles.scss
+++ b/src/scss/styles.scss
@@ -406,6 +406,7 @@ input[type="search"]::-webkit-search-cancel-button {
         width: 45px;
         max-width: 45px;
         text-align: center;
+        line-height: 0;
 
         img {
             @extend .rounded;
@@ -417,6 +418,7 @@ input[type="search"]::-webkit-search-cancel-button {
     td.table-list-checkbox {
         width: 35px;
         max-width: 35px;
+        line-height: 0;
     }
 
     td.table-list-strike {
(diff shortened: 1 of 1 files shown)


src/scss/styles.scss
```diff
@@ -740,6 +742,11 @@ app-user-billing {
     }
 }
 
+.form-check-label {
+    margin: 0.3rem 0 0 0.3rem;
+    line-height: 16px;
+}
```
kspearrin, line 748: We use bootstrap for these styles. Maybe you should propose these fixes upstream?

SrdjanMilic: For the icons and check boxes align, `line-height: 1.5;` is bootstrap property declared on body element that is making the mess. From my experience, those are the common Bootstrap inconsistencies. Please, check out the difference at last row (maybe not best example):

2020-05-04-18-21-vault bitwarden com

By the way, icons vertical alignment at `secure notes` mobile app section are also a mess. Really looks bad.  

When checking the form label `Remember email` styles on live site, it's seems that changes does not have the same effect as in my local app version. Please, take a look at my local version with changes:

Screenshot_2020-05-04 Bitwarden Web Vault

Live site with same changes in dev tools:

2020-05-04-19-02-vault bitwarden com

> Maybe you should propose these fixes upstream?

Sorry, but I can't figure it out what do you mean? I will do what ever it takes.

SrdjanMilic:
> By the way, icons vertical alignment at secure notes mobile app section are also a mess. Really looks bad.

Looking at desktop app, now I can see, that it is because an empty space reserved for note details. That is why they look as they are not aligned. Anyway, that looks ugly. Should be some condition to change classes when there is no details.

I couldn't find the way to add that details to secure note. How I can do that?

PR #532, opened 2020-05-15
fixing a few bug, asset updates, tweaks

kspearrin:
1. Moved timeout action change warning to when the user clicks the option rather than waiting until they press save button.
2. Fix a null reference error on add/edit page.
3. Fix localization string for bulk perm delete.
4. Fix missing styling on bulk restore button.
5. Show "shared" icon even when icon is deleted in trash listing. @cscharf Can you confirm we show this icon in trash for other clients too?
6. Update static assets for new branding.


cscharf:
>Can you confirm we show this icon in trash for other clients too?

I can, this display is unique for the web vault, the other clients don't seem to have "shared" indicators on the View page, only the share/shared info on the edit pages.

kspearrin:
@cscharf This shared icon change was on the listing page, not the add/edit or view pages.

cscharf:
My bad, yes, looking at the other clients it's only toggling on org vs. not (so your correction was correct, good catch).

PR #539, opened 2020-05-21
[Paging] Added for Organization Users, Pages, and Collections

vincentsalucci:
## Code Changes
- Added `infiniteScroll` UI components
- Added logic to handle loading more list items - follows precedent set by the `ciphers.component`
- Set page size to 100 for Users, Groups, and Collections

#### Deleting item last "page"
paging-delete

### DIFF
--- a/src/app/organizations/manage/collections.component.ts
+++ b/src/app/organizations/manage/collections.component.ts
@@ -14,6 +14,7 @@ import { ApiService } from 'jslib/abstractions/api.service';
 import { CollectionService } from 'jslib/abstractions/collection.service';
 import { I18nService } from 'jslib/abstractions/i18n.service';
 import { PlatformUtilsService } from 'jslib/abstractions/platformUtils.service';
+import { SearchService } from 'jslib/abstractions/search.service';
 import { UserService } from 'jslib/abstractions/user.service';
 
 import { CollectionData } from 'jslib/models/data/collectionData';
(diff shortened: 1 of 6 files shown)


src/app/organizations/manage/collections.component.ts
```diff
@@ -147,10 +171,28 @@ export class CollectionsComponent implements OnInit {
...
+            this.resetPaging();
+        }
+        return !searching && this.collections.length > this.pageSize;
+    }
+
     private removeCollection(collection: CollectionView) {
         const index = this.collections.indexOf(collection);
         if (index > -1) {
             this.collections.splice(index, 1);
+            this.resetPaging();
```
cscharf, line 195: Would we want to completely reset paging to the beginning when a collection is deleted? Does this reset the view in that sense so if I scroll through 5 pages, delete an item, I have to scroll back through those same 5 pages again to get back to where I was at?

src/app/organizations/manage/groups.component.ts
```diff
@@ -142,10 +165,28 @@ export class GroupsComponent implements OnInit {
...
+            this.resetPaging();
+        }
+        return !searching && this.groups.length > this.pageSize;
+    }
+
     private removeGroup(group: GroupResponse) {
         const index = this.groups.indexOf(group);
         if (index > -1) {
             this.groups.splice(index, 1);
+            this.resetPaging();
```
cscharf, line 189: Same comment here, would removing a group reset the user back to the beginning of their paging/scrolling?

src/app/organizations/manage/collections.component.ts
vincentsalucci: Negative - the `resetPaging` is basically used to catch up the paging list with the master list. Deleting a collection near the end of your list, has the same visible affect as deleting a collection at the top of your list.

src/app/organizations/manage/groups.component.ts
vincentsalucci: Same as above!

PR #542, opened 2020-05-25
Changed CSS to allow for wider screens

syntax-error752:


syntax-error752:
Will make further changes before opening a PR.

PR #543, opened 2020-05-26
Update CSS to allow for larger screens.

syntax-error752:
Allows for the scaling of the centre column in the vault.
**Before:**
before
**After:**
after

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -35,6 +35,7 @@ import { CryptoService } from 'jslib/abstractions/crypto.service';
 import { EventService } from 'jslib/abstractions/event.service';
 import { FolderService } from 'jslib/abstractions/folder.service';
 import { I18nService } from 'jslib/abstractions/i18n.service';
+import { MessagingService } from 'jslib/abstractions/messaging.service';
 import { NotificationsService } from 'jslib/abstractions/notifications.service';
 import { PasswordGenerationService } from 'jslib/abstractions/passwordGeneration.service';
 import { PlatformUtilsService } from 'jslib/abstractions/platformUtils.service';
@@ -69,6 +70,7 @@ export class AppComponent implements OnDestroy, OnInit {
     private lastActivity: number = null;
     private idleTimer: number = null;
     private isIdle = false;
+    private enableFullWidth = false;
 
     constructor(private angulartics2GoogleAnalytics: Angulartics2GoogleAnalytics,
         private broadcasterService: BroadcasterService, private userService: UserService,
@@ -83,7 +85,7 @@ export class AppComponent implements OnDestroy, OnInit {
         private sanitizer: DomSanitizer, private searchService: SearchService,
         private notificationsService: NotificationsService, private routerService: RouterService,
         private stateService: StateService, private eventService: EventService,
-        private policyService: PolicyService) { }
+        private policyService: PolicyService, private messagingService: MessagingService) { }
 
     ngOnInit() {
         this.ngZone.runOutsideAngular(() => {
@@ -148,6 +150,10 @@ export class AppComponent implements OnDestroy, OnInit {
                             properties: { label: message.label },
                         });
                         break;
+                    case 'setFullWidth':
+                        this.enableFullWidth = await this.storageService.get<boolean>('enableFullWidth');
+                        this.setFullWidth();
+                        break;
                     default:
                         break;
                 }
@@ -166,6 +172,7 @@ export class AppComponent implements OnDestroy, OnInit {
                 }
             }
         });
+        this.messagingService.send('setFullWidth');
     }
 
     ngOnDestroy() {
@@ -260,6 +267,14 @@ export class AppComponent implements OnDestroy, OnInit {
             this.notificationsService.disconnectFromInactivity();
         } else {
             this.notificationsService.reconnectFromActivity();
+        }
+    }
+
+    private async setFullWidth() {
+        if (this.enableFullWidth) {
+            document.body.classList.add('full-width');
+        } else {
+            document.body.classList.remove('full-width');
         }
     }
 }
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -35,7 +35,6 @@ import { CryptoService } from 'jslib/abstractions/crypto.service';
 import { EventService } from 'jslib/abstractions/event.service';
 import { FolderService } from 'jslib/abstractions/folder.service';
 import { I18nService } from 'jslib/abstractions/i18n.service';
-import { MessagingService } from 'jslib/abstractions/messaging.service';
 import { NotificationsService } from 'jslib/abstractions/notifications.service';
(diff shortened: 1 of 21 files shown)


src/scss/styles.scss
```diff
@@ -832,3 +833,16 @@ img.logo {
 .cursor-move {
     cursor: move !important;
 }
+
+.col-3 {
+    max-width: 240px;
```
cscharf, line 838: Bootstrap uses flex-box and `%` for controlling `col-*` widths in a fluid layout. I wouldn't want to cap those simply because of the bootstrap styles that are being used for the web vault's layout without more specificity. If there are regions/parts of the web vault's layout that need to be capped on width that are currently only styles as `col-3`, for example the **FILTERS** sidebar, that should then get an additional css class appended to the markup itself, to be targeted by the stylesheet specifically:
```less
.left-nav {
    max-width: 240px;
}
```
OR, the use of specificity, nesting and *nth* operators be used for targeting those elements, i.e.
```less
.container {
    /*...*/
    .row {
        .col-3 {
            &:first-child {
                max-width: 240px;
            }
        }
    }
}
```

src/scss/styles.scss
```diff
@@ -832,3 +833,16 @@ img.logo {
...
+}
+
+.col-5 {
+    max-width: 420px;
```
cscharf, line 842: Same

src/scss/styles.scss
```diff
@@ -832,3 +833,16 @@ img.logo {
...
+}
+
+.col-6 {
+    flex: 0 0 calc(100% - 500px);
```
cscharf, line 846: Same here, essentially ensuring specificity to the primary content region of the web vault layout instead of any `col-6` class'd element.

src/scss/styles.scss
syntax-error752: Changes made as requested.

syntax-error752:
> I'm not sure I fully see the value in stretching the site full-width (well 90% anyway) while keeping the filters and organizations sections fluid as well (vs. fixed), since a larger screen would better allow longer organization names, other languages, folder names/structure, etc. to prevent wrapping in larger view-ports, etc. Also, `980px` is a pretty decent size as-is, I'm not sure why a fully fluid layout would be necessary for the web vault since most if not all use-cases would fail to make use of the expanded screen real-estate (can you provide a screen shot or example of where this would be more appropriate or a better UX?)

In our small business we have recently started using Bitwarden, and have imported many on the entries into our self-hosted vault.
A few people here, including myself find it a bit annoying that the vault view is limited to a small column, while all the (poorly named) entries are caused to wrap, and the space to the side gets wasted.
Ideally we would rename these, but here is a small selection of some of our long names (imported from the old password manager), copied into an otherwise empty vault (for security reasons):
usage_1
Compared to expanding the columns:
usage_2

I have implemented this is our self-hosted environment, and in our case at least, people seem to like the changes, if only because it makes the long (and poorly) named entries easier to read.

src/scss/styles.scss
```diff
@@ -188,12 +188,25 @@ input, select, textarea {
 }
 
 .container {
-    width: 980px;
+    min-width: 980px;
```
cscharf, line 191: Please move these 2 lines inside another class wrapper, `.full-width`, e.g.:
```less
.container {
    width: 980px;
    &.full-width {
        min-width: 980px;
        width: 90%;
    }
}
```

src/scss/styles.scss
```diff
@@ -188,12 +188,25 @@ input, select, textarea {
...
+    width: 90%;
     max-width: none !important;
     margin: 0 auto;
     padding: 0;
+    .row {
```
cscharf, line 196: I would think this should be removed... leaving the `.col-*` classes as-is should still allow full-width display to work w/o wrapping and we would want scaling to work across the board (not just for the center region).

src/scss/styles.scss
```diff
@@ -188,12 +188,25 @@ input, select, textarea {
...
+            max-width: 420px;
+        }
+        .col-6 {
+            flex: 0 0 calc(75% - 240px);
+            max-width: none;
+        }
+    }
 }
-
+
```
cscharf, line 209: Please remove added whitespace

cscharf:
@syntax-error752 , please provide localized strings only in English as part of the primary English language file, we use Crowdin upon commit/merge to `master` to source translations for those strings once published.

syntax-error752:
I have converted this to an option:
image

The scaling does not affect the top nav bar or footer.

cscharf:
Awesome! 2 small requests:
* Please ensure the navigation bar also scales (kind of looks funny being in the center-ish of the page otherwise)
* Please label the option: "Full Width Layout", the description underneath is perfect.

src/app/app.component.ts
```diff
@@ -166,6 +172,7 @@ export class AppComponent implements OnDestroy, OnInit {
                 }
             }
         });
+        this.messagingService.send('setFullWidth');
```
syntax-error752, line 175: Added to trigger on load.

src/services/htmlStorage.service.ts
```diff
@@ -4,9 +4,9 @@ import { ConstantsService } from 'jslib/services';
 
 export class HtmlStorageService implements StorageService {
     private localStorageKeys = new Set(['appId', 'anonymousAppId', 'rememberedEmail', 'passwordGenerationOptions',
-        ConstantsService.disableFaviconKey, 'rememberEmail', 'enableGravatars', ConstantsService.localeKey,
```
syntax-error752, line 7: Had to rearrange  to keep within the line character limit.

src/scss/styles.scss
```diff
@@ -123,6 +123,18 @@ body {
...
     }
+
+    &.full-width {
+        .container {
+            min-width: 980px;
+            width: 90%;
+            .row {
+                .col-5 {
+                    max-width: 416px;
+                }
```
syntax-error752, line 134: col-5 max-width set to keep the login box the same size regardless of weather on not the fill width setting is turned on.

syntax-error752:
The requested changes have been made.

src/scss/styles.scss
cscharf: Let's remove the override for the login form itself, if we need to reset it I would prefer it get it's own "functional class" for css targeting (e.g. "`.login-form`") but looking at it, while it's certainly a lot wider in this scenario, I believe it's acceptable without limiting its size.

src/scss/styles.scss
```diff
@@ -123,6 +123,18 @@ body {
     &.layout_frontend {
         background-color: #ecf0f5;
     }
+
+    &.full-width {
```
kspearrin, line 127: Just add `:not(&.layout_frontend)` or something to that affect to exclude the frontend layout.

syntax-error752:
Updated to use the not selector (with a pre-existing class).

PR #550, opened 2020-06-04
stub alternate layout support for register page

kspearrin:
This stubs out the layout pages for marketing team


src/app/accounts/register.component.html
```diff
@@ -1,101 +1,127 @@
 <form #form (ngSubmit)="submit()" [appApiAction]="formPromise" class="container" ngNativeValidate>
```
kspearrin, line 1: Man, GitHub is terrible at showing diffs

PR #558, opened 2020-06-12
Feature/tax info collection

cscharf:
## Objective

Collect tax information for the following situations:

1. User is adding a credit to their personal account
2. User is upgrading to premium, entering payment details or using a credit
3. User is changing their billing payment method
4. User is creating a new organization without a "Free" plan
5. Organization's billing payment method is being changed
6. Organization is updating their Tax Information directly from their Settings page
7. Organization is adding a credit to their account

## Changes

New TaxInfo component, re-used anywhere we collect tax information.
Added Country + Postal Code (ZIP required for US)
For Orgs, added checkbox for including VAT/GST and Address info

## Screenshots

image

image

image


cscharf:
I'll update jslib and resolve conflicts once jslib PR is approved, https://github.com/bitwarden/jslib/pull/115

src/app/organizations/settings/account.component.html
```diff
@@ -49,9 +49,13 @@ <h1>{{'apiKey' | i18n}}</h1>
     <h1>{{'taxInformation' | i18n}}</h1>
 </div>
 <p>{{'taxInformationDesc' | i18n}}</p>
-<a href="https://bitwarden.com/contact/" target="_blank" rel="noopener" class="btn btn-outline-secondary">
-    {{'contactSupport' | i18n}}
-</a>
+<form *ngIf="org && !loading" #formTax (ngSubmit)="submitTaxInfo()" [appApiAction]="formPromise" ngNativeValidate>
```
kspearrin, line 52: If we're not going to show anything in this section while things are loading, I would recommend showing a spinner in some way instead of nothing.

src/app/organizations/settings/account.component.ts
```diff
@@ -71,6 +73,13 @@ export class AccountComponent {
         } catch { }
     }
 
+    async submitTaxInfo() {
+        this.formPromise = this.taxInfo.submitTaxInfo();
```
kspearrin, line 77: You need a new promise variable. Don't share it with another form. I think this would activate its spinners and such. Create `taxFormPromise` or something.

src/app/settings/premium.component.ts
```diff
@@ -76,6 +78,7 @@ export class PremiumComponent implements OnInit {
                     return this.finalizePremium();
                 });
             } else {
+                await this.taxInfoComponent.submitTaxInfo();
```
kspearrin, line 81: I would prefer we include the tax info in a single request like we do with org create, with postPremium here. Looking at your server code, I don't see how this would even properly work for a User who does not have a Stripe customer record yet. That gets created in postPremium. I think that the same issue exists in other places where you are calling two separate API calls, like add Credit.

cscharf: Yeah, I had almost gone that route, but there was a lot of back and forth trying to minimize duplicative code and reusable components. I'll refactor the server and then ensure these follow suit to append/tack-on the tax info to any calls that would normally use it; while keeping the independent "save" there for organizations from their settings page (outside of billing).

cscharf: **// TODO/:**

(each includes UI, jslib and server changes)
1. `add-credit.component.ts` > Need to refactor server method to accept tax info, append tax-info to add-credit collective data/JSON
2. `adjust-payment.component.ts` > Add tax info to the `PaymentRequest` type (similar to `organization-plans.component.ts`
3. `premium.component.ts` > Add tax info to `FormData` to submit for `postPremium(FormData)`

kspearrin: For add credit, that may need to stay a separate step since I don't think we make a call to our server during that action. Credit is added asynchronously via webhook.

cscharf: Yeah, I'm seeing that... I'll check to see if those services enable state transfer or a custom blob/field that I can encode and pass that info across and then use from the web-hook. But otherwise have to kind of rely on them already having a Stripe customer id, otherwise it would fail/be ignored. 

This _may_ not be necessary at all since I'm assuming we're not realizing that revenue (and therefore no tax) when a credit is added and **only** when that credit is applied to an invoice for premium subscription services would we count that ARR/revenue. In that same light, the modification planned for `premium.component.ts` will handle requiring tax information anyway so we may not need this at all when adding account credit since we will always have the billing tax info when actually invoicing (either the first time or from subsequent subscription renewal invoices in Stripe).

PR #562, opened 2020-06-25
Added referenceId to register component

cscharf:
## Objective

Capture and pass referenceId to the server on registration.

## Changes

* **Register Component** - set referenceId from passed in query string OR cookie

### DIFF
--- a/src/app/accounts/register.component.ts
+++ b/src/app/accounts/register.component.ts
@@ -79,6 +79,11 @@ export class RegisterComponent extends BaseRegisterComponent {
             }
             if (qParams.reference != null) {
                 this.referenceId = qParams.reference;
+            } else {
+                this.referenceId = ('; ' + document.cookie).split('; reference=').pop().split(';').shift();
+            }
+            if (this.referenceId === '') {
+                this.referenceId = null;
             }
             if (queryParamsSub != null) {
                 queryParamsSub.unsubscribe();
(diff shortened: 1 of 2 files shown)


cscharf:
Will update jslib and create full PR once jslib done/merged.

src/app/accounts/register.component.ts
```diff
@@ -96,6 +99,12 @@ export class RegisterComponent extends BaseRegisterComponent {
         if (this.policies != null) {
             this.enforcedPolicyOptions = await this.policyService.getMasterPasswordPolicyOptions(this.policies);
         }
+
+        const referenceId = ('; ' + document.cookie).split('; reference=').pop().split(';').shift();
+        this.referenceId = this.referenceId || referenceId;
+        if (this.referenceId === '') {
+            this.referenceId = null;
+        }
```
kspearrin, line 107: Does this need to be moved into the queryParams subscribe callback? That is async

cscharf: Yes it does, and this has now been moved, verified after some testing that even with no query parameters this callback is fired from the queryParams observable, before ngOnInit completes so I bet the observable fires immediately but is still async, moved code to prevent a race condition.

src/app/layouts/organization-layout.component.ts
```diff
@@ -8,8 +8,8 @@ import { ActivatedRoute } from '@angular/router';
 
 import { BroadcasterService } from 'jslib/angular/services/broadcaster.service';
 
-import { UserService } from 'jslib/abstractions/user.service';
 import { EnvironmentService } from 'jslib/abstractions/environment.service';
+import { UserService } from 'jslib/abstractions/user.service';
```
cscharf, line 12: This resolved a lint warning (which are really, really annoying ;-)

PR #575, opened 2020-07-15
SSO support

kspearrin:
- sso component to kick off sso auth flow
- sso connector to catch redirect callback from sso flows


src/app/accounts/login.component.html
```diff
@@ -44,6 +44,11 @@
                             <i class="fa fa-pencil-square-o" aria-hidden="true"></i> {{'createAccount' | i18n}}
                         </a>
                     </div>
+                    <div class="d-flex">
```
cscharf, line 47: Should this be hidden for now from the UI?

kspearrin: We can circle back and comment it out if we want to remove for a release.

PR #578, opened 2020-07-17
Transition reference id to data

cscharf:
## Objective
Ensure we pass layout, flow and reference id in the new reference data parameter when registering a new user. Will break until jslib is updated.


src/app/accounts/register.component.ts
```diff
@@ -19,6 +19,7 @@ import { MasterPasswordPolicyOptions } from 'jslib/models/domain/masterPasswordP
 import { Policy } from 'jslib/models/domain/policy';
 
 import { PolicyData } from 'jslib/models/data/policyData';
+import { ReferenceEventData } from 'jslib/models/domain/referenceEventData';
```
kspearrin, line 22: per other PRs, path will change to API request model

cscharf: Fixed

PR #581, opened 2020-07-23
Prevent password autocomplete on edit

bilhackmac:
Prevent browser password suggestion and current password overriding in edit mode

### DIFF
--- a/src/app/vault/add-edit.component.html
+++ b/src/app/vault/add-edit.component.html
@@ -71,7 +71,7 @@
                             <div class="input-group">
                                 <input id="loginPassword" class="form-control text-monospace"
                                     type="{{showPassword ? 'text' : 'password'}}" name="Login.Password"
-                                    [(ngModel)]="cipher.login.password" appInputVerbatim autocomplete="new-password"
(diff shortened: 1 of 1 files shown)


kspearrin:
Please provide reproduction steps. `new-password` should already prevent autofilling.

bilhackmac:
Under macOS (10.15.4) with Safari (13.1) simply connect to your vault and edit any entry with password.

Select password field and then iCloud Keychain automatically suggest new password.
At this point, you can accept suggested password or ignore it. But if you hit ignore button, field become empty.

But strangely this happen only once by vault unlock.

eliykat:
I'm unable to reproduce this bug - I only get a small, unobtrusive key icon when editing a password in Safari:

Screen Shot 2021-05-25 at 4 01 38 pm

We also haven't received any other reports of this issue. It's possible that it's been resolved by Apple or by us in the meantime. On this basis I'm going to close this PR.

@bilhackmac, if you're still experiencing this bug and you can provide steps to reliably reproduce it, please comment here. I'll take a look and reopen if needed.

PR #585, opened 2020-07-27
Fix modal-body div missing #583

### LINKED ISSUE #583: modal-body div missing when inviting user to organisation
## Describe the Bug
When clicking the button to invite a user to an organisation, a modal is displayed containing a modal-head, modal-body and modal-footer div. This modal-body div contains information relating to the user you're inviting (email address, user type etc.). However, if the list of current people in the organisation is filtered to either "Invited" or "Accepted" when you click the Invite User button, the modal-body div is missing, making the modal essentially empty.

## Screenshots or Videos
How modal should look
bw-invite-user

How it looks when current people filtered to Invited or Accepted
bw-invite-user-empty

## Environment

- Windows 10 Pro
- Edge Version 84.0.522.40 

## Additional Context
I'm using a free organisation where I can share with just one other person

cscharf:
Fixes #583 

This was originally introduced by #539 (paging) which caused, on initial render, the `this.users`, `this.collections` and/or `this.groups` arrays to not yet be initialized (were `undefined` or `null` and therefore caused JavaScript exceptions). This didn't impact other functionality because they would later become initialized and paging would work appropriately. However, if you clicked on a category/filter that was empty, and then tried to click "Add/Invite", etc., it would cause a re-render/init call based on Angular's 2-way bindings (for `isPaging()` and `reload()`) and these arrays would once again be `null` and/or `undefined` causing an exception and failing to render the component in the modal body once launched.

### DIFF
--- a/src/app/organizations/manage/collections.component.ts
+++ b/src/app/organizations/manage/collections.component.ts
@@ -85,7 +85,7 @@ export class CollectionsComponent implements OnInit {
     }
 
     loadMore() {
-        if (!this.collections || this.collections.length <= this.pageSize) {
+        if (this.collections.length <= this.pageSize) {
             return;
         }
         const pagedLength = this.pagedCollections.length;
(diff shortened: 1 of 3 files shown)


cscharf:
I am trying to figure out how to do that w/ jslib... that change wasn't queued in my commit or staged changes when I did the commit and pushed, and when I try to revert it's not there either, so I'm at a bit of a loss. I'll get that figured out here shortly 🤞

PR #598, opened 2020-08-11
Price and Plan Updates

addisonbeck:
This is a (mostly) identical PR to what was approved already in web-internal. It has been updated to use latest jslib.


src/app/settings/organization-plans.component.html
```diff
@@ -209,8 +210,16 @@ <h2 class="spaced-header mb-4">{{'paymentInformation' | i18n}}</h2>
         </ng-container>
         <small class="text-muted font-italic mt-2 d-block" *ngIf="!createOrganization">
             {{'paymentCharged' | i18n : (interval | i18n) }}</small>
-    </ng-container>
-    <div [ngClass]="{'mt-4': !createOrganization || plans[plan].noPayment}">
+    </div>
+    <div class="mt-4">
```
addisonbeck, line 214: I am wondering if we should toggle this off until an API is chosen and implemented? Right now it just shows on screen that sales tax is 0$.

src/app/settings/organization-plans.component.html
```diff
@@ -209,8 +210,16 @@ <h2 class="spaced-header mb-4">{{'paymentInformation' | i18n}}</h2>
...
+        <div id="price" class="my-4" *ngIf="total > 0">
+            <div class="text-muted text-sm">Plan price: {{ subtotal | currency: 'USD $' }}
+                <br />Estimated tax: {{ salesTax | currency: 'USD $' }}</div>
```
addisonbeck, line 217: I am wondering if we should toggle this off until an API is chosen and implemented? Right now it just shows on screen that sales tax is 0$.

src/app/settings/organization-plans.component.ts
```diff
@@ -100,6 +64,137 @@ export class OrganizationPlansComponent {
...
+        }
+        if (this.selectedPlan.hasPremiumAccessOption && this.premiumAccessAddon) {
+            total += this.selectedPlan.premiumAccessOptionPrice;
+        }
+        return total;
+    }
+
+    // TODO: Wire up to chosen API
+    get salesTax() {
+        return 0;
```
cscharf, line 154: Missed this before... we actually want to use our our API for calculating sales tax... then any changes really happen server-side for us, not here in the UI. It would be our API that would return `0` for the time being (this would actually call the API with the necessary parameters to calculate sales tax, and then only for US).

src/app/settings/organization-plans.component.ts
```diff
@@ -100,6 +64,137 @@ export class OrganizationPlansComponent {
...
+    }
+
+    get total() {
+        return this.subtotal + this.salesTax || 0;
```
cscharf, line 158: Can we please make the order of operations here explicit?
```typescript
return this.subtotal + (this.salesTax || 0);
```

src/app/settings/organization-plans.component.ts
```diff
@@ -100,6 +64,137 @@ export class OrganizationPlansComponent {
...
+    }
+
+    changedOwnedBusiness() {
+        if (!this.ownedBusiness || this.selectedPlan.canBeUsedByBusiness) {
+            return;
+        }
+        this.plan = PlanType.TeamsMonthly;
+    }
+
+    changedCountry() {
```
cscharf, line 184: this method can also control sales tax re-calculation OR at least visibility of the Sales Tax element.

src/app/settings/organization-plans.component.html
cscharf: This needs to be visible only if the selected billing country is the US. Otherwise we would leave it, even if it's always zero.

Also, we should be using our own API for doing this, so the UI would never know when the server is going to start returning a value other than zero.

src/app/settings/organization-plans.component.ts
addisonbeck: I don't think this method can control re-calculation, that would make sales tax calculations go stale and be inaccurate with seat/storage/plan changes. 

I am concerned about the network usage that goes in to recalculating sales tax, especially considering we are going to our server and then going to whatever API is selected for computation. The are a lot of fields impacting sales tax amount. Recalculation is going to be needed on pretty much every change event in the form. Should we consider moving the subtotal/tax/total information to a separate confirmation screen?

cscharf: @addisonbeck , Fair enough, which is why sales tax was a stretch goal, let's scrap it for now and we'll revisit after SSO and Plan Updates are delivered/live! Thanks Addison!

src/app/settings/organization-plans.component.html
addisonbeck: Removed

src/app/settings/organization-plans.component.ts
addisonbeck: Removed

src/app/settings/organization-plans.component.ts
addisonbeck: Removed

src/app/settings/organization-plans.component.ts
addisonbeck: Took it all out

src/app/settings/organization-plans.component.html
```diff
@@ -140,7 +140,7 @@ <h2 class="spaced-header">{{'summary' | i18n}}</h2>
                 <ng-container *ngIf="selectablePlan.isAnnual">
                     {{'annually' | i18n}}
                     <small *ngIf="selectablePlan.basePrice">
-                        {{'basePrice' | i18n}}: {{selectablePlan.basePrice / 12 | currency:'$'}} &times;12
+                        {{'basePrice' | i18n}}: {{selectedPlanInterval | currency:'$'}} &times;12
```
cscharf: If this is `"Annual"` or `"Monthly"`, I'm not seeing how that translates into a base price or a currency.

addisonbeck: Yeah, I had gotten a little carried away with that property refactor. Already fixed in latest.

PR #604, opened 2020-08-13
SSO feature

kspearrin:
Merging sso feature branch


cscharf:
If no other commits to merge, looks good.

PR #605, opened 2020-08-14
Modifications made to support Browser Extension SSO

MattTheDev:
## Objective

Add support for Browser Extension SSO. Receive initial request, transition to IDP, and return state/code to extension for final processing.

## Code Changes 

1. Include ConstantsService in sso.component.cs extension to provide access to stored objects via their constants
2. Override ngOnInit to provide web specific SSO flow.


src/app/accounts/sso.component.ts
```diff
@@ -31,4 +32,23 @@ export class SsoComponent extends BaseSsoComponent {
...
+                    window.postMessage({ type: "AUTH_RESULT", code: qParams.code, state: qParams.state }, "*");
+                }
+            }
+
+            if (qParams.state.endsWith(':clientId=browser')) {
+                this.redirectUri = qParams.redirectUri;
+                this.state = qParams.state;
+                this.clientId = qParams.clientId;
+                await this.storageService.save(ConstantsService.ssoStateKey, qParams.state);
+            }
```
kspearrin, line 49: still a little confused why this is needed.

src/app/accounts/sso.component.ts
```diff
@@ -31,4 +32,23 @@ export class SsoComponent extends BaseSsoComponent {
...
         this.clientId = 'web';
     }
+
+    async ngOnInit() {
+        const queryParamsSub = this.route.queryParams.subscribe(async (qParams) => {
+            if (qParams.code != null && qParams.state != null) {
+                if (qParams.state.endsWith(':clientId=browser')) {
+                    window.postMessage({ type: "AUTH_RESULT", code: qParams.code, state: qParams.state }, "*");
+                }
+            }
```
kspearrin, line 42: just had another thought. Could we just do this in the sso-connector.html file instead of needing to redirect back into this sso component?

src/app/accounts/sso.component.ts
MattTheDev: BaseSsoComponent isn't holding onto a value that's needed. If this isn't there .. instead of continuing the flow and redirecting to the extension, it redirects to the Org ID enter screen and stops.

Let me see if I can find the exact values and see if there is a way to handle them.

src/app/accounts/sso.component.ts
MattTheDev: Are you thinking in a script block? @kspearrin

kspearrin: Yes, there is JS there already that is parsing the query params

src/app/accounts/sso.component.ts
kspearrin: state will come back in your redirect params. you should not need to save it.

MattTheDev: Removed. Seems to work fine without this.

src/app/accounts/sso.component.ts
MattTheDev: Like that approach. Tested and submitted PR.

MattTheDev:
Waiting on #606 so I can merge it in to fix the build.

PR #606, opened 2020-08-14
Upgrade Angular to 9

Hinton:
Updated using `ng update`.

Depends on https://github.com/bitwarden/jslib/pull/151


src/app/tools/cipher-report.component.ts
```diff
@@ -15,8 +16,9 @@ import { AddEditComponent } from '../vault/add-edit.component';
 import { MessagingService } from 'jslib/abstractions/messaging.service';
 import { UserService } from 'jslib/abstractions/user.service';
 
+@Directive()
```
kspearrin, line 19: I am a little confused why this is only on some components.

Hinton: My understanding is that in Angular 9, all classes are required to have either the `Directive` or `Component` decorator. Since this "component" is missing a template file, it became a `Directive` in the migration.

https://angular.io/guide/migration-undecorated-classes

PR #609, opened 2020-08-17
call api to set password with key parameters

kspearrin:
The existing set password flow did not work due to missing keys and other crypto related proeprties on the user account which are not set yet. This upadte extends the change password component to support the set password flow for new sso users.

### DIFF
--- a/src/app/accounts/change-password.component.ts
+++ b/src/app/accounts/change-password.component.ts
@@ -19,19 +19,25 @@ import { UserService } from 'jslib/abstractions/user.service';
 import { CipherString } from 'jslib/models/domain/cipherString';
 import { SymmetricCryptoKey } from 'jslib/models/domain/symmetricCryptoKey';
 
+import { KeysRequest } from 'jslib/models/request/keysRequest';
 import { SetPasswordRequest } from 'jslib/models/request/setPasswordRequest';
 
 import {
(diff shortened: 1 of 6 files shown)


src/app/accounts/change-password.component.ts
```diff
@@ -19,19 +19,25 @@ import { UserService } from 'jslib/abstractions/user.service';
...
     selector: 'app-accounts-change-password',
     templateUrl: 'change-password.component.html',
 })
 export class ChangePasswordComponent extends BaseChangePasswordComponent {
+    showPassword: boolean = false;
+    hint: string = '';
+
     onSuccessfulChangePassword: () => Promise<any>;
-    successRoute = 'lock';
+    successRoute = 'vault';
```
vincentsalucci, line 40: Can you explain why this is changing?

kspearrin: Since we are already collecting the master password during this step, we can go ahead and set the needed keys to avoid the extra unlock step.

vincentsalucci: Okay - just thought we agreed on having them re-enter it anyway (similar to how when you change a password you have to re-enter anyway)

kspearrin: Yea, I think we originally did say that. After implementing this part of it, I realized we could optimize that flow a bit.

PR #610, opened 2020-08-17
Upgrade Angular CDK

Hinton:
Forgot to upgrade this dependency when upgrading to Angular 9.


kspearrin:
Why is it a different version than all the others still?

Hinton:
Angular CDK follows the Angular Material Design version numbers. There is no `9.1.12` available, and 9.2.4 supports everything on 9.1.

PR #611, opened 2020-08-17
ESlint & Prettier

Hinton:
Since TSlint is deprecated this PR changes the linter to ESlint which is the recommended replacement along with Prettier for code formatting.

At the same time I added Husky and lint-staged to automatically run linting pre-commit. It uses --fix command and will only fail if it cannot fix the error.

Most of the custom TSlint settings have been moved over, a few was discarded since they either no longer exists. I disabled a few of the recommended rules since they played badly with the code base and generated many errors. While they are solvable they would require a fair bit of time since they cannot be auto-fixed.

For Prettier I modified the default line-width to 100 as 80 produces a lot more changes.


Hinton:
I noticed there are a few tslint:disable-next-line, will remove them.

src/app/organizations/tools/exposed-passwords-report.component.ts
```diff
@@ -1,17 +1,12 @@
...
 
 import { AuditService } from 'jslib/abstractions/audit.service';
 import { CipherService } from 'jslib/abstractions/cipher.service';
 import { MessagingService } from 'jslib/abstractions/messaging.service';
 import { UserService } from 'jslib/abstractions/user.service';
 
-import {
-    ExposedPasswordsReportComponent as BaseExposedPasswordsReportComponent,
-} from '../../tools/exposed-passwords-report.component';
+import { ExposedPasswordsReportComponent as BaseExposedPasswordsReportComponent } from '../../tools/exposed-passwords-report.component';
```
kspearrin, line 9: What about these formatting cases? This is breaking the 120 max line rule... ?

Hinton: Prettier has no explicit max line rule. It will make a best effort to strive to use a specific line length but makes no promise that it will do so if the alternative is sub-optimal.

.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+    "singleQuote": true,
+    "printWidth": 100,
```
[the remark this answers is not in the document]
Hinton: https://prettier.io/docs/en/options.html

PR #650, opened 2020-09-14
only use memory storage for vault data keys

kspearrin:
Recently I changed this storage service to use memory storage instead of session storage to address the limitations with session storage lengths. This was probably a bit too aggressive since the only keys were data grows large is in the vault. This change adds checks for only using memory storage for specific vault data keys so that other keys can still use session storage. This prevents things like the user being logged out completely on page refresh.


cscharf (approved):
Good call.

Akash-Hamal:
Thanks for quick response and i hope that it will be fixed soon! Thanks Team!

PR #654, opened 2020-09-15
653 - fix user agent detection for Edge

### LINKED ISSUE #653: "New Device Logged In From Chrome" email when logging in from Edge
When logging in from latest stable Edge I get an email telling me about a new login from Chrome

cscharf:
Fixes and closes #653 

## Overview
This is caused by the Edge browser detection from the user-agent string provided by the Edge browser, which Microsoft changed to use the `Edg/` token vs. the `Edge/` token previously used to help differentiate over confusion from the prior, Trident based IE Edge browser.

### DIFF
--- a/src/services/webPlatformUtils.service.ts
+++ b/src/services/webPlatformUtils.service.ts
@@ -24,7 +24,7 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
             this.browserCache = DeviceType.FirefoxBrowser;
         } else if (navigator.userAgent.indexOf(' OPR/') >= 0) {
             this.browserCache = DeviceType.OperaBrowser;
-        } else if (navigator.userAgent.indexOf(' Edge/') !== -1 || navigator.userAgent.indexOf(' Edg/') !== -1) {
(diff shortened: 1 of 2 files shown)


src/services/webPlatformUtils.service.ts
```diff
@@ -24,7 +24,7 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
             this.browserCache = DeviceType.FirefoxBrowser;
         } else if (navigator.userAgent.indexOf(' OPR/') >= 0) {
             this.browserCache = DeviceType.OperaBrowser;
-        } else if (navigator.userAgent.indexOf(' Edge/') !== -1) {
+        } else if (navigator.userAgent.indexOf(' Edge/') !== -1 || navigator.userAgent.indexOf(' Edg/') !== -1) {
```
kspearrin, line 27: By doing this, you will introduce a bunch of hacks that are in place for Edge device type. We need to strip those out since we will no longer be deploying for old edge.

kspearrin (approved):

cscharf:
Updated jslib

kspearrin (approved):

kspearrin (approved):

Issue #567, opened 2020-07-03
[Security][UX] Password hint field on sign up too close to password field


mcfedr:
## Describe the Bug

I was doing some maintenance on my self hosted Bitwarden instance, and whilst looking in the `User` table, I noticed quite a few `MasterPasswordHint`s that look suspiciously like a password, and on further inspect, turn out to be the actual password of the user.

## Steps To Reproduce

Hard to say exactly, I wonder if the non-english speaking nature of my users has an effect on this, but clearly something in the UX isn't making it clear what should be in that box

## Expected Result

Hint, should be a hint, not the password.

I'm thinking some validation that this isnt the case is required.

I would question how helpful the hint really is...

## Actual Result

Hint contains the users password.

## Screenshots or Videos

<img width="512" alt="Screenshot 2020-07-03 at 3 49 50 pm" src="https://user-images.githubusercontent.com/704356/86470933-17eca800-bd45-11ea-9c99-c65e90354dba.png">

## Environment

- Build Version: [2.15.1] - but most of these users have signed up before this.

cwille97:
I'm going to work on a fix for this tomorrow that will reject any hints that are the same as the password since that isn't very secure.

djsmith85:
This issue can be closed. The initial PR #685 was closed due to another PR for the jslib component https://github.com/bitwarden/jslib/pull/194

PR #685, opened 2020-10-30
issue567 Prevent users from using their password as the password hint

cwille97:
It's not very secure to allow the user to use their actual password as the hint, and it seems like some users that do not speak the language are confused by the hint prompt and have been typing their password into it. I think the UI itself is straightforward enough and shouldn't cause confusion, but I think we should prevent the user from actually submitting a hint that matches the master password. Link to issue: https://github.com/bitwarden/web/issues/567#issuecomment-719014993

### DIFF
--- a/src/app/accounts/register.component.ts
+++ b/src/app/accounts/register.component.ts
@@ -117,6 +117,7 @@ export class RegisterComponent extends BaseRegisterComponent {
                 this.i18nService.t('masterPasswordPolicyRequirementsNotMet'));
             return;
         }
+        this.verifyPasswordhint();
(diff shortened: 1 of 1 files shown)


kspearrin:
This should probably be in the base register component in jslib so that all projects can implement it.

cwille97:
@kspearrin I updated the PR, please see the corresponding PR on the jslib repo: https://github.com/bitwarden/jslib/pull/194

cwille97:
This should no longer be necessary. The verification is moved to the `submit()` function which is already being called here.

PR #691, opened 2020-11-07
Add 'Copy Username' button

djmattyg007:
This adds a 'Copy Username' button above the 'Copy Password' button in the dropdown for individual entries in the safe. This matches the capabilities of the desktop app, where you can right-click on any entry and get options for both 'copy password' and 'copy username'.

Currently the only way to easily copy the username is to manually highlight the name under the entry name. Double-clicking or triple-clicking to highlight the entire username often isn't suitable either, because it usually results in extraneous whitespace being copied.


cscharf (approved):

cscharf:
Thanks @djmattyg007 ! Nice UX improvement.

Issue #666, opened 2020-10-04
Issue in local developement and remote server data fetch


exploit-io:
Hi,

I am making some changes in web version, when i run `npm run build:prod:watch` The web page loads in https://localhost:8080

But when I try to login with my creds, It cant fetch data from bitwarden API.

In the browser it says the error is because of CORS!

Please help me solve the problem.

exploit-io:
I could solve the problem with following extension.


https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf?hl=en

PR #706, opened 2020-11-15
Updated readme to note how to run the app against production

jdmccauley:
Hi Bitwarden, 

I am working on a feature for this repo, and I ran into some issue running the [app against production](https://community.bitwarden.com/t/catch-and-prompt-user-of-potential-duplicate-passwords-on-import/14955/14). I found solutions to my issues ([major one here](https://github.com/bitwarden/web/issues/666)), and I wanted to update the README so others don't run into the same issues.

Changes:

- Noted Node.js version can only be up to 13.9.0 because of SCSS version support

- Added events url to reflect how services.module.ts actually looks

- Added instructions for how to use the correct npm command for running the app against production

Please let me know what you think about these changes, and if there is anything I should add or remove.

Best,
Josh

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
 
 ### Requirements
 
-- [Node.js](https://nodejs.org) v8.11-v13.9.0
+- [Node.js](https://nodejs.org) v8.11 or greater
 
 ### Run the app
 
--- a/README.md
+++ b/README.md
@@ -63,6 +63,20 @@ npm install
 npm run build:prod:watch
 ```
 
+## Common Issues:
+
+### CORS
+
+If you run the frontend and receive a notification after attempting to login that says:
+```
+An error has occurred.
+NetworkError when attempting to fetch resource.
+```
+And in the console:
+```
+Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://api.bitwarden.com/accounts/prelogin. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,10 +23,12 @@
 
 ### Requirements
 
-- [Node.js](https://nodejs.org) v8.11 or greater
+- [Node.js](https://nodejs.org) v8.11-v13.9.0
```
cscharf, line 26: node-sass does provide support all the way up to nodejs v15+; not sure the specific issue you ran into on this one, but support does vary for that package from version to version (including minor and patch releases) for sure. I would leave this as ` or greater`.

cscharf (commented):
One comment, thanks for the update/submission, I'm sure this will help some folks. Did you also want to add anything you found in regards to the CORS issues when running against production?

README.md
jdmccauley: Okay that makes sense, I just ran into a dependency issue when I've been trying to build. I had to install nodejs v13 since it had issues with my original nodejs v15. Unsure why, but I agree let's keep it as is.

jdmccauley:
Done, I hope the CORS issue mention helps. What do you think?

README.md
cscharf: Can you revert the change back to ` or greater` please? Then I think this will be ready to go!

cscharf (commented):
Just that one change to the requirement version of node.js needed and then this should be ready to go. The CORS section looks great, thank you!

jdmccauley:
Done, thanks for catching that!

cscharf (approved):
awesome, thanks!

PR #708, opened 2020-11-18
send service implementation

kspearrin:
Implement jslib's new send service through send components.


kspearrin:
I think I need to still update jslib ref after that is merged.

cscharf (approved):

PR #719, opened 2020-11-24
WIP: dirty fix to SSO web vs browser redirect logic split

MGibson1:
# Objective

Currently, SSO sign on from the browser extension signs into web. It should open a webpage to allow for specification of SSO settings, complete SSO, and redirect to the extension for vault decryption.

# Files Changed

* **sso.ts**: state string was changed to include the organization identifier, but this broke the pivot we were using for web vs browser sign-ons. This is a quick fix for this, but it may be that a contains is better suited or even moving/removing identifier.  I don't really have the expertise in the feature to understand the implecations of changing the state string. It's clearly pretty fragile.

### DIFF
--- a/src/connectors/sso.ts
+++ b/src/connectors/sso.ts
@@ -5,7 +5,7 @@ document.addEventListener('DOMContentLoaded', (event) => {
     const code = getQsParam('code');
     const state = getQsParam('state');
 
-    if (state != null && state.endsWith(':clientId=browser')) {
+    if (state != null && state.endsWith(':clientId=browser_identifier=sso')) {
         initiateBrowserSso(code, state);
     } else {
         window.location.href = window.location.origin + '/#/sso?code=' + code + '&state=' + state;
--- a/src/connectors/sso.ts
+++ b/src/connectors/sso.ts
@@ -5,7 +5,7 @@ document.addEventListener('DOMContentLoaded', (event) => {
     const code = getQsParam('code');
     const state = getQsParam('state');
 
-    if (state != null && state.endsWith(':clientId=browser_identifier=sso')) {
+    if (state != null && state.includes(':clientId=browser')) {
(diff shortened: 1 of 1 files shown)


src/connectors/sso.ts
```diff
@@ -5,7 +5,7 @@ document.addEventListener('DOMContentLoaded', (event) => {
     const code = getQsParam('code');
     const state = getQsParam('state');
 
-    if (state != null && state.endsWith(':clientId=browser')) {
+    if (state != null && state.endsWith(':clientId=browser_identifier=sso')) {
```
addisonbeck, line 8:
```suggestion
    if (state != null && state.includes(':clientId=browser')) {
```

Organization identifier is a user-set value, so we can't use it explicitly in an endsWith. I'd just check the clientId parameter standalone with an includes.

addisonbeck:
Does this work? Your comment in Slack about the experience still being bad sounds like something is still be broken.

If you're just concerned about the magic-string nature of this whole feature, take heart, because SSO is built on magic strings 😁

src/connectors/sso.ts
MGibson1: ah good point! that's what happens when you tunnel vision your test case.

MGibson1:
@addisonbeck, includes is definitely better, don't know how I missed that.

As for the experience, the biggest hurdle I have right now is the page you enter your organization ID in hangs open and I haven't figured out how to close it yet.

cscharf:
> @addisonbeck, includes is definitely better, don't know how I missed that.
> 
> As for the experience, the biggest hurdle I have right now is the page you enter your organization ID in hangs open and I haven't figured out how to close it yet.

I do not think that's possible, however it's worth a look because we have the same issue with Desktop and CLI.

cscharf (approved):

PR #720, opened 2020-11-25
Update web sso content to indicate window OK to close

MGibson1:
This is done after the authResult handoff message is delivered to the
extension. It is not possible to close the window from javascript as
closing a window is limited to the script that opened it.

If we maintain a reference to the web window, it should be possible to
subscribe to the authResult message and close the web windows from the
browser.

# Overview
updates the web sso logon page to indicate it is OK to close.

# Image
image

### DIFF
--- a/src/app/accounts/sso.component.ts
+++ b/src/app/accounts/sso.component.ts
@@ -53,6 +53,9 @@ export class SsoComponent extends BaseSsoComponent {
 
     async submit() {
         await this.storageService.save(IdentifierStorageKey, this.identifier);
+        if (this.clientId === 'browser') {
+            document.cookie = `ssoHandOffMessage = ${this.i18nService.t('ssoHandOff')}`
+        }
         super.submit();
     }
 }
--- a/src/app/accounts/sso.component.ts
+++ b/src/app/accounts/sso.component.ts
@@ -54,7 +54,7 @@ export class SsoComponent extends BaseSsoComponent {
     async submit() {
         await this.storageService.save(IdentifierStorageKey, this.identifier);
         if (this.clientId === 'browser') {
-            document.cookie = `ssoHandOffMessage = ${this.i18nService.t('ssoHandOff')}`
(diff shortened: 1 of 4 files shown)


src/connectors/sso.ts
```diff
@@ -37,6 +37,8 @@ function getQsParam(name: string) {
 
 function initiateBrowserSso(code: string, state: string) {
     window.postMessage({ command: 'authResult', code: code, state: state }, '*');
+    document.getElementById('content').innerHTML = '<p>Successfully passed off authentication to the extension.</p>' +
```
mportune-bw, line 40: I think the first line should just be "Authentication successful"

cscharf, line 40: We would need to figure out how to get the i18n service for localization in here to fetch this string value for multi-lingual support.

Also, perhaps we can re-word this to something like: 
> Authentication was successful. You may now close this tab and return to the browser extension.

cscharf (commented):
Is authentication successful in this scenario still (browser extension is actually logged in)?

src/connectors/sso.ts
addisonbeck, line 40: This string should get run through the the translation process. To simply it so you only need one tag (and one messages string) I'd recommend something shorter like "Success! You are now logged in to the extension".

addisonbeck: Gonna just resolve this since it was already said

src/app/accounts/sso.component.ts
```diff
@@ -53,6 +53,9 @@ export class SsoComponent extends BaseSsoComponent {
 
     async submit() {
         await this.storageService.save(IdentifierStorageKey, this.identifier);
+        if (this.clientId === 'browser') {
+            document.cookie = `ssoHandOffMessage = ${this.i18nService.t('ssoHandOff')};SameSite=strict`
```
cscharf, line 57: Should the space around the `=` be removed?, i.e. `ssoHandOffMessage=$this...`

cscharf (approved):

PR #733, opened 2020-12-09
Add jslib to npm dependency tree

MGibson1:
# Overview

Currently bitwarden/jslib is referenced only as a submodule and any new dependency in that repo needs to be manually added to all 4 client repositories that depend on jslib. I'm adding jslib as a local package dependency to npm, which includes it in dependency tree calculations.

# Files Changed

* **package.json**: jslib dependency addition.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,6 +24,7 @@ install:
     $env:PUSH_DOCKER = "false"
     $env:PROD_DEPLOY = "false"
     $env:TAG_NAME = ""
+    git submodule update --init --recursive
     if($env:APPVEYOR_REPO_TAG -eq "true" -and $env:APPVEYOR_RE_BUILD -eq "True") {
       $env:PROD_DEPLOY = "true"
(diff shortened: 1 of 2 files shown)


cscharf (changes requested):
```
npm install
npm ERR! code ENOLOCAL
npm ERR! Could not install from "jslib" as it does not contain a package.json file.
npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\appveyor\AppData\Roaming\npm-cache\_logs\2020-12-09T15_49_06_007Z-debug.log
Command exited with code 1
```

MGibson1:
yeah should have made draft (and browser one) to test out building.

Looks like an issue with appveyor config

MGibson1:
It looks like this is just a bad idea in this case. jslib is requiring packages to be installed on our ubuntu image in order to build, which is just the worst.

It's a decision between breaking our CI pipelines and adding a package during jslib update PRs. I think adding a package is the friendlier solution. Sorry for the ultimately useless pinging.

PR #735, opened 2020-12-09
Add ConsoleLogService dependency from jslib

MGibson1:
# Overview

Use the default consoleLogService. jslib is moving toward requiring clients to specify a console logger. This change also includes the `browser-process-hrtime` dependency

# File changes

* **packages.json**: Add new dependency
* **services.ts**: Add new required ConsoleLogService definition

### DIFF
--- a/src/app/services/services.module.ts
+++ b/src/app/services/services.module.ts
@@ -30,6 +30,7 @@ import { AuditService } from 'jslib/services/audit.service';
 import { AuthService } from 'jslib/services/auth.service';
 import { CipherService } from 'jslib/services/cipher.service';
 import { CollectionService } from 'jslib/services/collection.service';
+import { ConsoleLogService } from 'jslib/services/consoleLog.service';
 import { ConstantsService } from 'jslib/services/constants.service';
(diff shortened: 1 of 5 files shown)


src/app/services/services.module.ts
```diff
@@ -84,6 +84,7 @@ import { TokenService as TokenServiceAbstraction } from 'jslib/abstractions/toke
 import { TotpService as TotpServiceAbstraction } from 'jslib/abstractions/totp.service';
 import { UserService as UserServiceAbstraction } from 'jslib/abstractions/user.service';
 import { VaultTimeoutService as VaultTimeoutServiceAbstraction } from 'jslib/abstractions/vaultTimeout.service';
+import { ConsoleLogService } from 'jslib/services/consoleLog.service';
```
cscharf, line 87: Please move this import up into the `jslib/services/**` block above with the other service implementation imports.

cscharf (changes requested):
Please see comments on imports

cscharf (approved):

PR #739, opened 2020-12-15
[Policy] Single Org dependency chain

vincentsalucci:
## Code Changes
- **policy-edit.component**: Added explicit check for when saving the `SingleOrg` policy to check for the above condition.
- **messages**: Added new error string

## Screenshots
0-web-error

### DIFF
--- a/src/app/organizations/manage/policy-edit.component.ts
+++ b/src/app/organizations/manage/policy-edit.component.ts
@@ -186,6 +186,18 @@ export class PolicyEditComponent implements OnInit {
                 }
                 return true;
 
+            case PolicyType.SingleOrg:
+                if (this.enabled) { // Don't need prevalidation checks if submitting to enable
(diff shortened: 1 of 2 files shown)


src/app/organizations/manage/policy-edit.component.ts
```diff
@@ -186,6 +186,18 @@ export class PolicyEditComponent implements OnInit {
                 }
                 return true;
 
+            case PolicyType.SingleOrg:
+                if (this.enabled) { // Don't need prevalidation checks if submitting to enable
```
cscharf, line 190: Let's move the code comment up above the `if (...)` statement vs. next to it.

vincentsalucci: Moved.

cscharf (approved):

PR #741, opened 2020-12-16
update jslib and fix webPlatformUtils

cscharf:
Resolved the API abstraction change from https://github.com/bitwarden/jslib/pull/161 for web.

### DIFF
--- a/src/services/webPlatformUtils.service.ts
+++ b/src/services/webPlatformUtils.service.ts
@@ -285,4 +285,12 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
     supportsSecureStorage() {
         return false;
     }
+
+    getDefaultSystemTheme() {
+        return null as 'light' | 'dark';
+    }
+
+    onDefaultSystemThemeChange() {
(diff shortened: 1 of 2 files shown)


src/services/webPlatformUtils.service.ts
```diff
@@ -285,4 +285,12 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
     supportsSecureStorage() {
         return false;
     }
+
```
MGibson1, line 288: Should we include some comments or something that this is stubbed out until we want to implement it?

vincentsalucci (approved):

MGibson1 (commented):
seems like this is a push till later type PR?

src/services/webPlatformUtils.service.ts
cscharf: not sure comments are really needed, this is filling in for the contract... when we get around to supporting themes in web it should be pretty straightforward. That's why `null` is a bit more explanatory IMO.

PR #410, opened 2019-09-03
Only show carats for items with children

DarkArc:
Adaptation of https://github.com/bitwarden/desktop/pull/303 for the web vault.

PR #745, opened 2020-12-18
show caret or icon, not just "blank" space

cscharf:
Does a little bit of reversal on the removal of carets from *all* items (see PR https://github.com/bitwarden/web/pull/410), and instead uses an icon for the item OR a caret (if expandable) for folders and collections. Needed to remove some of the formatting based on the `.carets` class as well in this instance because we need things to align with the stuff higher up.

image

### DIFF
--- a/src/app/vault/groupings.component.html
+++ b/src/app/vault/groupings.component.html
@@ -62,7 +62,7 @@
                         <i class="fa fa-plus fa-fw" aria-hidden="true"></i>
                     </a>
                 </h3>
-                <ul class="fa-ul card-ul carets">
+                <ul class="fa-ul card-ul">
                     <ng-template #recursiveFolders let-folders>
                         <li *ngFor="let f of folders"
                             [ngClass]="{active: selectedFolder && f.node.id === selectedFolderId}">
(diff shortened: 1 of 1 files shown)


MGibson1:
This could just be my eyes, but to me it's appears that `Matrioska Doll` is a child of `Got Folders?` I think I'd like 
(caret or blank) **and** icon the best. Takes up a fair amount of horizontal space, though.

cscharf:
> This could just be my eyes, but to me it's appears that `Matrioska Doll` is a child of `Got Folders?` I think I'd like
> (caret or blank) **and** icon the best. Takes up a fair amount of horizontal space, though.

I toyed with a lot of options here... I agree with the original poster (and it has caught me multiple times) that when we show the caret on every folder whether it has sub-folders or not, it's confusing and I end up clicking things for clicking sake. It's also misleading for a non-visual user because the screen reader will tell you there are descendants and an expand/collapse, but then no children would be read after moving to an expanded state.

When the original requestor's changes were implemented, however, for longer vaults or where there were no nested folders, you just see oddly margin'd folder names w/o anything and it looks pretty wonky.

Even when you do have a couple of nested folders w/ carets, the rest of the folders still look wonky.

I ended up going with an icon **or** caret approach, using visual nesting (so the starting "M" and icon for `Matrioska Doll` is aligned with the folder above, not nested.) It also removes the expansion nuance for non-visual users and as a non-focusable element for keyboard users. It's a trade-off but I couldn't find a better visual representation. Optional caret + icon could work, but we would need to really tool with the styles and layout so we're not squishing text for the sake of iconography.

MGibson1 (approved):

PR #747, opened 2020-12-21
fixing incorrect secret name

joseph-flinn:
## Change Summary
Fixing incorrectly named secret in workflow

## Files Changed
- .github/workflows/build.yml


cscharf (approved):
It was so secret, the name wasn't even known. Now it's less secret, lol

PR #757, opened 2020-12-30
send improvements and bug fixes

kspearrin:
- Move remove password and delete functions to send service, which properly handles updating server and local data
- Status icons on send listing page
- New string for disabled checkbox
- Max access count cannot be set to less than 1
- Better error messages for 404 and other errors when accessing send
- Remove auto-text selection on text send access

### DIFF
--- a/src/app/send/access.component.html
+++ b/src/app/send/access.component.html
@@ -25,6 +25,12 @@
                         </button>
                     </div>
                 </div>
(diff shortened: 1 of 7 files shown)


cscharf:
@kspearrin , did you want to push the latest from `jslib` once that's merged so this builds successfully in a single commit?

kspearrin:
@cscharf  done

cscharf (approved):

PR #758, opened 2020-12-30
Hide match uri overflow

### LINKED ISSUE #531: Can't delete URI
When I edit a password in the web vault, I can't delete URIs I don't need anymore.

grafik

MGibson1:
# Overview

Closes #531. Cipher view modal's URI match detection selector is not handling overflow properly and pushing the URI delete button off the modal.

This change handles overflow properly. Selecting the modal still displays the full text.

# Files Changed

# Screenshots

## pre change
image

## post change
image

### DIFF
--- a/src/app/vault/add-edit.component.html
+++ b/src/app/vault/add-edit.component.html
@@ -164,7 +164,7 @@
                                     </a>
                                 </div>
                                 <div class="d-flex">
-                                    <select class="form-control" id="loginUriMatch{{i}}" name="Login.Uris[{{i}}].Match"
+                                    <select class="form-control overflow-hidden" id="loginUriMatch{{i}}" name="Login.Uris[{{i}}].Match"
                                         [(ngModel)]="u.match" (change)="loginUriMatchChanged(u)" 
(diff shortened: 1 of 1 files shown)


src/app/vault/add-edit.component.html
```diff
@@ -164,7 +164,7 @@ <h2 class="modal-title" id="cipherAddEditTitle">{{title}}</h2>
                                     </a>
                                 </div>
                                 <div class="d-flex">
-                                    <select class="form-control" id="loginUriMatch{{i}}" name="Login.Uris[{{i}}].Match"
+                                    <select class="form-control overflow-hidden" id="loginUriMatch{{i}}" name="Login.Uris[{{i}}].Match"
```
MGibson1, line 167: It might be better to add overflow: hidden to all selectors rather than ad-hoc to problematic ones.

cscharf: Maybe, if it happens to another or becomes a problem we can do that then I think.

cscharf (approved):

PR #768, opened 2021-01-07
ui updates for send add/edit component

kspearrin:
- Added descriptions under most fields to give more context on what each Send option does on add/edit.
- Added client messages about premium being required when trying to create sends of type "File".
- "Text" type is default whenever not premium.

### DIFF
--- a/src/app/send/add-edit.component.ts
+++ b/src/app/send/add-edit.component.ts
@@ -12,6 +12,7 @@ import { SendType } from 'jslib/enums/sendType';
 
 import { EnvironmentService } from 'jslib/abstractions/environment.service';
 import { I18nService } from 'jslib/abstractions/i18n.service';
+import { MessagingService } from 'jslib/abstractions/messaging.service';
 import { PlatformUtilsService } from 'jslib/abstractions/platformUtils.service';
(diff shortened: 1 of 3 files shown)


src/app/send/add-edit.component.ts
```diff
@@ -10,19 +10,18 @@ import { Component } from '@angular/core';
...
+import { UserService } from 'jslib/abstractions/user.service';
 
 import { SendView } from 'jslib/models/view/sendView';
 import { SendFileView } from 'jslib/models/view/sendFileView';
 import { SendTextView } from 'jslib/models/view/sendTextView';
 
 import { Send } from 'jslib/models/domain/send';
-
-import { SendData } from 'jslib/models/data/sendData';
+import { MessagingService } from 'jslib/abstractions/messaging.service';
```
addisonbeck, line 24: Might want to move this import up with the other abstractions

addisonbeck (changes requested):

src/app/send/add-edit.component.ts
kspearrin: Done

addisonbeck (approved):

PR #775, opened 2021-01-14
fixed various Permissions UI issues

addisonbeck:
Fixed various UI issues with Permissions reported by @clayadams5226 during testing

1. If someone has a custom permission, when the admin goes to the user access screen (Manage > Collections > Users) Custom doesn't show up next to the user. It's just blank
2. If I have the policies permission, but not the portal permission, I'm still able to access the portal from the policies page (although I can't make any changes.)
3. The SSO permission is displayed as "Sso" should it all be caps?

### DIFF
--- a/src/locales/en/messages.json
+++ b/src/locales/en/messages.json
@@ -3611,7 +3611,7 @@
     "message": "Manage Policies"
   },
   "manageSso": {
-    "message": "Manage Sso"
+    "message": "Manage SSO"
   },
   "manageUsers": {
     "message": "Manage Users"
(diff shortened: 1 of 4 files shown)


src/locales/en/messages.json
```diff
@@ -3611,7 +3611,7 @@
     "message": "Manage Policies"
   },
   "manageSso": {
-    "message": "Manage Sso"
+    "message": "Manage SSO"
```
cscharf, line 3614: 👍🏻 , yes, it should be all caps

cscharf (approved):

PR #779, opened 2021-01-18
make inline buttons accessible with tab button

### LINKED ISSUE #712: Elements are not Accessible with the Keyboard #UX #Ergonomics #Accessibiliy
## Describe the Bug
In the Web Vault I checked the tab order and if all elements are accessible with the TAB key.

### Unlock Vault (enter master password)
Here all elements are accessible with the TAB key, **no issues were found**.

### My Vault
Here all elements are accessible with the TAB key, **no issues were found**.

### Adding / Editing an entry

image

The elements marked with a red border are not accessible with the TAB key.
Number 5 and number 6 focus the generate and check icons above the field.

At custom password fields also the buttons to see the password and to copy it to the clipboard are not accessible using the TAB key:

image

I know that I can copy the content of a field using CTRL+C, but nevertheless all UI-elements that are visible, enabled and interactive must be accessible with the keyboard.

### Tools
In the Tools I checked every single page, but also here all elements are accessible with the TAB key, **no issues were found**.

### Settings
Also in the Settings I checked every single page (I did **not yet check the Organizations**). 
Also here all elements are accessible with the TAB key, **no issues were found**.

## Steps To Reproduce
Just use the TAB key to cycle through the elements

## Expected Result
All interactive elements must be accessible with the keyboard (according to WCAG 2.1, chapters [2.1.1](https://www.w3.org/TR/WCAG21/#keyboard) and [2.1.2](https://www.w3.org/TR/WCAG21/#no-keyboard-trap))

## Environment
- Operating system: Windows 10 Pro 64-Bit Version 20H2 (Build 19042.630)
- Browser: Microsoft Edge Chromium 87.0.664.41 (Offizielles Build) (64-Bit)
- Build Version: 2.17.0

eliykat:
## Objective

Fix #712 - some buttons on the add-edit page were not accessible when tabbing through elements. The issue correctly identified that this is required by web accessibility guidelines. (See issue for relevant links)

## Code changes
Remove `tabindex=-1` from the relevant elements.

## Screenshots

I have tested this in Chrome and all elements now appear to be accessible via tab, including those that were raised in the issue report.

tab a11y

### DIFF
--- a/src/app/vault/add-edit.component.html
+++ b/src/app/vault/add-edit.component.html
@@ -47,7 +47,7 @@
                                 <div class="input-group-append" *ngIf="!cipher.isDeleted">
                                     <button type="button" class="btn btn-outline-secondary"
                                         appA11yTitle="{{'copyUsername' | i18n}}"
-                                        (click)="copy(cipher.login.username, 'username', 'Username')" tabindex="-1">
+                                        (click)="copy(cipher.login.username, 'username', 'Username')">
                                         <i class="fa fa-lg fa-clone" aria-hidden="true"></i>
                                     </button>
                                 </div>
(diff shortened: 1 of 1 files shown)


cscharf (approved):
Looks good, let's hold on this until after the release [code-freeze]

cscharf:
Go for merge.

PR #780, opened 2021-01-18
Clearer keyboard focus on input elements

### LINKED ISSUE #713: Keyboard Focus is not (or not clearly) visible #UX #Ergonomics #Accessibiliy
## Describe the Bug
In the Web Vault I checked if the keyboard focus is clearly visible on all focusable elements.
This issue is related to #712 (Elements are not Accessible with the Keyboard).

### Unlock Vault (enter master password)
The keyboard focus is not visible at all at the "Show Password" icon:
image

At the button "Logout" the keyboard focus is visible, but it is not clearly visible:
image

Please use a blue border (like you have it on other elements):
image

### Enter 2FA Code
The focus on the "Cancel" button is not clearly visible:
image

### My Vault
In My Vault the keyboard focus is not clearly visible at the "Options" button:
image

Please use a blue border, like you have it on the "Add Item" button.
image

Same for the "Options" button that is shown behind each entry in my vault (focus is not clearly visible):
image

At all other elements the keyboard **focus is clearly visible**.

### Adding / Editing an entry
When adding/editing an entry, the keyboard focus is not visible at all at the following elements:

Icon "Copy Verification Code": 
image

Icon "Remove URI":
image

Icons "Remove" and "Drag to sort item":
image

Icon "Favorite":
image

At the following icons the keyboard focus is not displayed correctly (the bottom line of the focus is missing):
image
image

At the "Cancel" button the keyboard focus is not clearly visible:
image

### Tools
In the tools I discovered that the keyboard focus on the menu items is not always displayed as rectangle.
At the first menu item the focus is **displayed correctly**:
image

At the second menu entry, the focus is only shown at the left side and the right side:
image

Here the bottom line of the focus indicator is missing (this problem occurs on most of the following menu items):
image

In the password generator the keyboard focus is not clearly visible at the buttons "Copy Password" and "Password History":
image
image
Also here a blue frame instead of a grey frame wold indicate the focus much better.

### Settings
In the Settings I checked every single page (I did **not yet check the Organizations**).
Also here you have the issue, that the keyboard focus in the navigation menu is often truncated (first entry is OK):
image
image
image

At the page "Premium membership" the keyboard focus is not clearly visible at the buttons "Reinstate Subscription" and "Download License":
image
image
Did I mention that a blue frame is better....? :wink:

At the page "Billing" the keyboard focus is not clearly visible at the buttons "Add Credit" and "Change Payment Method":
image
image

At the page "Two-step Login" there are also buttons having a grey focus indicator:
image
image
This occurs on all listed providers.

And last at the page "Domain Rules" there is also a button with a grey focus indicator:
image

## Steps To Reproduce
Just use the TAB key to cycle through the elements

## Expected Result
The keyboard focus must be clearly visible on all interactive elements (according to WCAG 2.1, chapters [2.4.7](https://www.w3.org/TR/WCAG21/#focus-visible) and [3.2.1](https://www.w3.org/TR/WCAG21/#on-focus)).

## Environment
- Operating system: Windows 10 Pro 64-Bit Version 20H2 (Build 19042.630)
- Browser: Microsoft Edge Chromium 87.0.664.41 (Offizielles Build) (64-Bit)
- Build Version: 2.17.0

eliykat:
## Objective

Fix #713: when tabbing through elements, not all elements have a clear outline/shadow to indicate that they have focus. This is required by WCAG.

## Code changes

I inspected each of the elements identified in the issue and grouped them as follows:

* `.btn-outline-secondary`: sets `box-shadow` on focus, but uses a colour with poor visibility. Changed so that it uses the primary colour for `box-shadow`.
* `.btn-link`: does not set any `box-shadow` or `outline` on focus. Changed to set `outline` on focus.
* `.list-group-item`: sets `outline` on focus, but this is partially obscured by other `.list-group-item`s nearby. Changed to set `z-index` on focus so that it's always on top.
* `<a>` items above text input: sets `outline` on focus, but this is partially obscured by the nearby text `input`. Changed to reduce the height of the element so the outline doesn't overlap.

I went through Bootstrap 4's default variables and mixins, but I don't think they let us customise this behaviour. Therefore this is just done by applying our own scss to these elements.

## Screenshots

`.btn-outline-secondary`
Screen Shot 2021-01-19 at 7 12 02 am

`.btn-link`
Screen Shot 2021-01-19 at 7 11 47 am

`.list-group-item`
Screen Shot 2021-01-19 at 7 12 32 am

`<a>` items above text input
Screen Shot 2021-01-19 at 7 12 54 am

### DIFF
--- a/src/app/vault/add-edit.component.html
+++ b/src/app/vault/add-edit.component.html
@@ -57,12 +57,12 @@
                             <div class="d-flex">
                                 <label for="loginPassword">{{'password' | i18n}}</label>
                                 <div class="ml-auto d-flex" *ngIf="!cipher.isDeleted && !viewOnly">
-                                    <a href="#" class="d-block mr-2" appStopClick
+                                    <a href="#" class="d-block mr-2 fa-icon-above-input" appStopClick
                                         appA11yTitle="{{'generatePassword' | i18n}}" (click)="generatePassword()"
                                         *ngIf="cipher.viewPassword">
                                         <i class="fa fa-lg fa-fw fa-refresh" aria-hidden="true"></i>
                                     </a>
-                                    <a href="#" class="d-block" #checkPasswordBtn appStopClick
+                                    <a href="#" class="d-block fa-icon-above-input" #checkPasswordBtn appStopClick
                                         appA11yTitle="{{'checkPassword' | i18n}}" (click)="checkPassword()"
                                         [appApiAction]="checkPasswordPromise">
                                         <i class="fa fa-lg fa-fw fa-check-circle" [hidden]="checkPasswordBtn.loading"
--- a/src/scss/styles.scss
+++ b/src/scss/styles.scss
@@ -315,6 +315,16 @@ input[type="search"]::-webkit-search-cancel-button {
     }
 }
 
+.btn-link {
+    &:focus,
+    &.focus {
+        outline-color: -webkit-focus-ring-color;
+        outline-offset: 1px;
+        outline-style: auto;
+        outline-width: 1px;
+    }
+}
+
 .btn-outline-secondary {
     color: $text-muted;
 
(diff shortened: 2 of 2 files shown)


cscharf (commented):
Looks good to me, @vincentsalucci , can you take a quick look and any final approval/remarks please?

vincentsalucci (approved):

PR #800, opened 2021-01-23
Add Emergency Access to Get Premium

vachan-maker:
Added "Emergency Access" to  src/app/settings/premium.component.html and messages.json

### DIFF
--- a/src/app/settings/premium.component.html
+++ b/src/app/settings/premium.component.html
@@ -16,6 +16,9 @@
             {{'premiumSignUpTwoStep' | i18n}}
         </li>
         <li>
+            <i class="fa fa-check text-success fa-il" aria-hidden="true"></i>
+            {{'premiumSignUpEmergency' |i18n}}
+        </li>
             <i class="fa fa-check text-success fa-li" aria-hidden="true"></i>
             {{'premiumSignUpReports' | i18n}}
         </li>
--- a/src/app/settings/premium.component.html
+++ b/src/app/settings/premium.component.html
@@ -16,7 +16,7 @@
             {{'premiumSignUpTwoStep' | i18n}}
         </li>
         <li>
-            <i class="fa fa-check text-success fa-il" aria-hidden="true"></i>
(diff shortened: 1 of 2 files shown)


src/app/settings/premium.component.html
```diff
@@ -16,6 +16,9 @@ <h1>{{'goPremium' | i18n}}</h1>
             {{'premiumSignUpTwoStep' | i18n}}
         </li>
         <li>
+            <i class="fa fa-check text-success fa-il" aria-hidden="true"></i>
+            {{'premiumSignUpEmergency' |i18n}}
+        </li>
```
Hinton, line 21: We are missing an opening tag here.

vachan-maker: Sorry, but I did not understand what you meant.
I have opened the li tags and closed it as well. Which other tag do I have to include?
Thanks

cscharf: @vachan-maker , you're missing the opening `<li>` tag for the next list item below your change. You have a closing `</li>` tag noted where Oscar added this note, but then if you look below, there's no opening `<li>` tag for the next closing `</li>` tag in the DOM for `{{'premiumSignUpReports' | i18n}}`

src/locales/en/messages.json
```diff
@@ -1565,6 +1565,9 @@
   "premiumSignUpTwoStep": {
     "message": "Additional two-step login options such as YubiKey, FIDO U2F, and Duo."
   },
+  "premiumEmergencyAccess": {
```
cscharf, line 1568: Sorry, just reviewing the rest of these changes now, looks like this key doesn't match what you're referencing in the html component, `premiumSignUpEmergency` vs. `premiumEmergencyAccess`; I can't find the key `premiumSignUpEmergency` anywhere in the current messages.json file.

vachan-maker: I have updated it. Sorry for making this complicated.

cscharf (approved):
No worries, thanks for the contribution!

PR #813, opened 2021-01-28
Set WebVaultUrl if dev

MGibson1:
# Overview

While playing with Sends in web, they were giving the wrong webvault URL. This change fixes that issue and should only impact development environments.

# Files Changed

* **services.module.ts**: Uses the default dev webvault url as we do in `apiService.SetUrls` just below this change.

# Testing concerns

This should have a minimal impact, but please test SSO needs to be validated as webVaultUrl is used for the return address from sso.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9941,6 +9941,14 @@
       "integrity": "sha1-J5siXfHVgrH1TmWt3UNS4Y+qBxM=",
       "dev": true
     },
+    "string_decoder": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-1.1.1.tgz",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -80,7 +80,7 @@
     "big-integer": "1.6.36",
     "bootstrap": "4.3.1",
     "braintree-web-drop-in": "1.13.0",
-    "browser-process-hrtime": "1.0.0",
+    "browser-hrtime": "^1.1.8",
```
kspearrin, line 83: need to run `npm i` to update lock file?

addisonbeck (approved):

addisonbeck:
Merged myself because @MGibson1 was experiencing strange github behavior

PR #818, opened 2021-02-03
Fix glob processing in npm. Ban single param parens

MGibson1:
# Overview

It turns out there was a bug in out npm linter script which allowed LOTS of lint violations through. This PR fixes that error.

It also attempts to standardize a lint ruleset across bitwarden typescript repositories. We'll likely need a little more work on this front, though.

# Files Changed
* **package.json**: single quotes around globs are resolved by shell rather than npm.
* **tslint.conf**: 
  * ordered imports is a big hassle for our review process (I'm sure you don't like having to do this manual @cscharf). This gets one step toward removing that burden. Eventually we will want to investigate the groups option and standardize how we group our imports as well.
  * do not allow parenthesis in arrow functions with single parameters otherwise parenthesis are required.

Other file changes are linter autofixes


cscharf (approved):
Yay! 😁

PR #819, opened 2021-02-04
Implement disable send policy

MGibson1:
# Overview

Depends on bitwarden/jslib#259

Implements web policy editor enabling of `DisableSend` policy. This policy disallows users of an organization from creating or editing Bitwarden Sends. Delete is still allowed.

# Files Changed

* **policies.component/policy-edit.component**: add DisableSend policy to the list of policies
* **send/add-edit.component**:  make all inputs either disabled or readonly (prefer readonly) if disable send policy is in effect. Also display an app-callout at the top of the modal and disable the save button. This had to be done with a second button due to the loading effect on the first being tied to the button's disabled state.
* **Send.component**: Add banner warning about sends being disabled. Disable the create button
* **messages.json**: messages relating to disable send policy

# Testing Concerns

Need to validate the disable logic from bitwarden/jslib#259. Send should be disabled when:
  * **Any** organization a user belongs to disables them
  * unless that user can manage policies by being
     1. Admin
     2. Owner
     3. custom with manage policy permission

Ensure polish of experience that disabled reason is not vague nor are you ever mistakenly thinking you're work can be saved.

# Screenshots

image

### Text send
image
image

### File send
image

### DIFF
--- a/src/app/send/send.component.ts
+++ b/src/app/send/send.component.ts
@@ -33,7 +33,8 @@ export class SendComponent extends BaseSendComponent {
 
     constructor(sendService: SendService, i18nService: I18nService,
         platformUtilsService: PlatformUtilsService, environmentService: EnvironmentService,
-        broadcasterService: BroadcasterService, ngZone: NgZone, searchService: SearchService, policyService: PolicyService, userService: UserService,
+        broadcasterService: BroadcasterService, ngZone: NgZone, searchService: SearchService,
+        policyService: PolicyService, userService: UserService,
         private componentFactoryResolver: ComponentFactoryResolver) {
         super(sendService, i18nService, platformUtilsService, environmentService, broadcasterService, ngZone,
             searchService, policyService, userService);
--- a/src/app/send/send.component.ts
+++ b/src/app/send/send.component.ts
@@ -19,6 +19,8 @@ import { PlatformUtilsService } from 'jslib/abstractions/platformUtils.service';
 import { SendService } from 'jslib/abstractions/send.service';
 
 import { BroadcasterService } from 'jslib/angular/services/broadcaster.service';
+import { PolicyService } from 'jslib/abstractions/policy.service';
+import { SearchService, UserService } from 'jslib/abstractions';
 
 @Component({
     selector: 'app-send',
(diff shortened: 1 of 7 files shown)


src/app/send/send.component.ts
```diff
@@ -31,12 +33,17 @@ export class SendComponent extends BaseSendComponent {
 
     constructor(sendService: SendService, i18nService: I18nService,
         platformUtilsService: PlatformUtilsService, environmentService: EnvironmentService,
-        broadcasterService: BroadcasterService, ngZone: NgZone,
+        broadcasterService: BroadcasterService, ngZone: NgZone, searchService: SearchService, policyService: PolicyService, userService: UserService,
```
cscharf, line 36: wrap `policyService: PolicyService, userService: UserService,` on a new line

src/locales/en/messages.json
```diff
@@ -3562,6 +3562,21 @@
   "personalOwnershipSubmitError": {
     "message": "Due to an Enterprise Policy, you are restricted from saving items to your personal vault. Change the Ownership option to an organization and choose from available Collections."
   },
+  "disableSend": {
+    "message": "Disable Send"
```
cscharf, line 3566: Each of these needs:
`"description": "'Send' is a noun and the name of a feature called 'Bitwarden Send'. It should not be translated."`

src/locales/en/messages.json
```diff
@@ -3562,6 +3562,21 @@
...
+  },
+  "disableSendPolicyDesc": {
+    "message": "Do not allow users to create or edit Bitwarden Sends. Deleting existing Sends is still allowed."
```
cscharf, line 3569: Let's stick with branding since "Send" is a proper noun and avoid plural "Send" if we can help it. Perhaps:
`"Do not allow users to create or edit a Bitwarden Send. Deleting an existing Send is still allowed."`

If we can't help it, it's fine, but you'll also need to allow variance in pluralization in the language for translation (adding a "s" doesn't work in most languages for making a proper noun plural, and a description of that would be difficult to convey to translators).

src/locales/en/messages.json
```diff
@@ -3562,6 +3562,21 @@
...
+  },
+  "disableSendExemption": {
+    "message": "Organization users that can manage the organization's policies are exempt from this policy's enforcement."
+  },
+  "sendDisabled": {
+    "message": "Send disabled"
+  },
+  "sendDisabledWarning": {
+    "message": "Due to an Enterprise Policy, you are restricted to only Deleting Sends."
```
cscharf, line 3578: Do we capitalize "Enterprise Policy" like this in other policy warning messages, same with the action Deleting. If we can say "deleting a Send" that would be awesome too IMO

MGibson1: I got the capitalization from personal ownership messaging, but it looks like single org doesn't capitalize. Let's decide which we like and I'll change all three

cscharf: Thanks for checking, let's go with non-capitalized everything except for "Send" which is a proper noun/brand

cscharf (approved):

PR #820, opened 2021-02-05
Improved handling of grantor access to organizations after takeover

eliykat:
This PR is required to support the changes described in https://github.com/bitwarden/server/pull/1132. Refer to that PR for explanation.

### DIFF
--- a/src/app/settings/emergency-access-takeover.component.ts
+++ b/src/app/settings/emergency-access-takeover.component.ts
@@ -19,7 +19,10 @@ import { UserService } from 'jslib/abstractions/user.service';
 import { ChangePasswordComponent } from 'jslib/angular/components/change-password.component';
 
 import { KdfType } from 'jslib/enums/kdfType';
+import { Policy } from 'jslib/models/domain/policy';
+import { PolicyData } from 'jslib/models/data/policyData';
 import { SymmetricCryptoKey } from 'jslib/models/domain/symmetricCryptoKey';
+import { EmergencyAccessTakeoverResponse } from 'jslib/models/response/emergencyAccessResponse';
 import { EmergencyAccessPasswordRequest } from 'jslib/models/request/emergencyAccessPasswordRequest';
 
 @Component({
@@ -35,6 +38,7 @@ export class EmergencyAccessTakeoverComponent extends ChangePasswordComponent im
     @Input() kdfIterations: number;
 
     formPromise: Promise<any>;
+    takeoverResponse: EmergencyAccessTakeoverResponse;
 
     constructor(i18nService: I18nService, cryptoService: CryptoService,
         messagingService: MessagingService, userService: UserService,
@@ -45,17 +49,19 @@ export class EmergencyAccessTakeoverComponent extends ChangePasswordComponent im
             platformUtilsService, policyService);
     }
 
-    // tslint:disable-next-line
-    async ngOnInit() { }
+    async ngOnInit() {
+        this.takeoverResponse = await this.apiService.postEmergencyAccessTakeover(this.emergencyAccessId);
+        const policyData = this.takeoverResponse.policy.map((policyResponse) => new PolicyData(policyResponse));
+        const policy = policyData.map((policyData) => new Policy(policyData));
+        this.enforcedPolicyOptions = await this.policyService.getMasterPasswordPolicyOptions(policy);
+     }
 
     async submit() {
         if (!await this.strongPassword()) {
             return;
         }
 
-        const takeoverResponse = await this.apiService.postEmergencyAccessTakeover(this.emergencyAccessId);
-
-        const oldKeyBuffer = await this.cryptoService.rsaDecrypt(takeoverResponse.keyEncrypted);
+        const oldKeyBuffer = await this.cryptoService.rsaDecrypt(this.takeoverResponse.keyEncrypted);
         const oldEncKey = new SymmetricCryptoKey(oldKeyBuffer);
 
         if (oldEncKey == null) {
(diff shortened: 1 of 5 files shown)


src/locales/en/messages.json
```diff
@@ -3417,6 +3417,9 @@
   "emergencyAccessDesc": {
     "message": "Grant and manage emergency access for trusted contacts. Trusted contacts may request access to either View or Takeover your account in case of a emergency. Visit our help page for more information and details into how zero knowledge sharing works."
```
cscharf, line 3418: While you're in here, can you please correct this one too?
> your account in case of a emergency.

Should be
> your account in case of **an** emergency.

cscharf (commented):
Some basic feedback, would love one of the other team members to take a look.

eliykat:
Thanks Chad! I'll wait for another review of this one before merging any of the other PRs, so they can all be merged at once.

cscharf (approved):
Actually, I think this looks good. Thanks Thomas!

cscharf (changes requested):
Actually, just noticed the build is failing. May need to take a look at that.

eliykat:
Except for an implicit any error (which I've just fixed), the build errors are only due to changes in jslib. Works fine if you have the updated jslib hooked up locally.

cscharf (approved):
that's better 😁

cscharf (changes requested):
nm, spoke too soon
```
ERROR in src/app/settings/emergency-access-takeover.component.ts:55:35 - error TS2339: Property 'policy' does not exist on type 'EmergencyAccessTakeoverResponse'.

55         if (this.takeoverResponse.policy != null && this.takeoverResponse.policy.length > 0)
                                     ~~~~~~
src/app/settings/emergency-access-takeover.component.ts:55:75 - error TS2339: Property 'policy' does not exist on type 'EmergencyAccessTakeoverResponse'.

55         if (this.takeoverResponse.policy != null && this.takeoverResponse.policy.length > 0)
                                                                             ~~~~~~
src/app/settings/emergency-access-takeover.component.ts:57:50 - error TS2339: Property 'policy' does not exist on type 'EmergencyAccessTakeoverResponse'.

57             const policy = this.takeoverResponse.policy.map((policyResponse: PolicyResponse) => new Policy(new PolicyData(policyResponse)));
                                                    ~~~~~~
```

eliykat:
The `policy` property of `EmergencyAccessTakeoverResponse` is added by changes to jslib here: https://github.com/bitwarden/jslib/pull/260/files

eliykat:
There you go 🥳

cscharf (dismissed):

src/app/send/send.component.html
```diff
@@ -107,7 +107,7 @@ <h1>
                                         {{'copySendLink' | i18n}}
                                     </a>
                                     <a class="dropdown-item" href="#" appStopClick (click)="removePassword(s)"
-                                        *ngIf="s.password">
+                                        *ngIf="s.password && !disableSend">
```
Hinton, line 110: Not sure whats happening here. It seems to have picked up a unrelated change? https://github.com/bitwarden/web/blob/master/src/app/send/send.component.html#L110 seems to already have this change.

cscharf: merge resolution more than likely, should be okay.

src/app/settings/emergency-access-takeover.component.ts
```diff
@@ -45,25 +50,29 @@ export class EmergencyAccessTakeoverComponent extends ChangePasswordComponent im
...
+        const oldKeyBuffer = await this.cryptoService.rsaDecrypt(this.takeoverResponse.keyEncrypted);
         const oldEncKey = new SymmetricCryptoKey(oldKeyBuffer);
 
         if (oldEncKey == null) {
             this.toasterService.popAsync('error', this.i18nService.t('errorOccurred'), this.i18nService.t('unexpectedError'));
             return;
         }
 
-        const key = await this.cryptoService.makeKey(this.masterPassword, this.email, takeoverResponse.kdf, takeoverResponse.kdfIterations);
+        const key = await this.cryptoService.makeKey(this.masterPassword, this.email, this.takeoverResponse.kdf, this.takeoverResponse.kdfIterations);
```
Hinton, line 75: Do we know what happens if the user we takeover managed to change their encryption key between the modal opening and the submit event being triggered? The modal could potentially be opened quite a bit longer than before.

cscharf: That's a good question Oscar.... @eliykat ?

eliykat: Good question! I just tested this situation and it didn't stop the takeover from going through or cause any unexpected behaviour. That said, I want to make sure I'm understanding the crypto-magic behind takeover before waving this through, so I'll have a chat with Oscar.

If it is a problem - or if we're just not comfortable with having stale `takeoverResponse` data client-side as a matter of best practice - the alternative would be to separate out the api calls, so that we get the grantor policies on `ngOn Init()` and move the `takeoverResponse` call back into `submit()`.

eliykat: Update: it's not really an issue if the grantor changes their master password while this modal is open, but if they rotate their encryption key, it has some... extremely bad outcomes. I have successfully nuked one (1) local testing vault to verify this.

I'll separate out the api calls as I mentioned above.

eliykat:
I've refactored this so we get the grantor's policies via a separate api call instead of messing with the `takeoverResponse` object. More details:
* I rebased a couple of times to revert earlier commits on this branch, and to avoid pulling in unrelated commits on master (like the one that Oscar flagged). Should ultimately make the history a little tidier.
* The latest changes that need reviewing are to `emergency-access-takeover.component.ts`
* server repo has a fresh PR here: TBA
* jslib repo has a fresh PR here: TBA

cscharf (approved):

PR #833, opened 2021-02-17
Updated PL translation

lukasz-pyrzyk:
"Adres rozliczeniowy" is misleading, because it usually means an official company address used by the accounting. In this field we expect an email address, so let's emphasize it.


cscharf:
Hi @lukasz-pyrzyk , thank you for the update, however all translations of Bitwarden are located in Crowdin, please see [this particular string here](https://crowdin.com/translate/bitwarden-web/58/en-pl?filter=basic&value=0#q=billingEmail) for translations/corrections and thank you again.

PR #841, opened 2021-02-22
Improve import error messages

eliykat:
## Objective

Improve import error messages so that users can identify problem items easier.

Errors are currently shown in the format below. 

Screen Shot 2021-02-22 at 6 09 28 pm

We say that this index number refers to the index of the CSV. It actually refers to the index of the item in the `importResult.ciphers[]`, `importResult.folders[]` or `importResult.collections[]` arrays (as applicable). For ciphers this will mostly correspond to the CSV index number, with some adjustments, but not always.

## Solution

One suggestion was to provide the line number in the import file. However, I think it's better to show some actual item details so that users don't have to dig through the import file to identify it. (This is also easier than trying to match the item object with the unparsed import file.)

A user also requested that the error is more "permanent" so that it can be copy/pasted to assist with troubleshooting. This will also avoid excessively long toast messages.

The improved error format is below. It shows the index number, type of item, item name, and error message.

Screen Shot 2021-02-22 at 6 14 22 pm

Note: We discussed automatically converting overlong notes to attachments, but that is out of scope for this PR. I wanted to have a generic method for handling server-side model validation errors first. This PR should handle validation errors for all fields on ciphers, folders and collections, not just notes.

## Code changes

* disable default error handling by `appApiAction`
* change toast message to a modal (this might benefit from some tweaking to make the `textarea` bigger)

This requires the following PR for `jslib`: https://github.com/bitwarden/jslib/pull/280

### DIFF
--- a/src/app/tools/import.component.html
+++ b/src/app/tools/import.component.html
@@ -248,7 +248,7 @@
         <label for="fileContents">{{'orCopyPasteFileContents' | i18n}}</label>
         <textarea id="fileContents" class="form-control" name="FileContents" [(ngModel)]="fileContents"></textarea>
     </div>
-    <button type="submit" class="btn btn-primary btn-submit" [disabled]="form.loading">
+    <button type="submit" class="btn btn-primary btn-submit" [disabled]="loading">
         <i class="fa fa-spinner fa-spin" title="{{'loading' | i18n}}" aria-hidden="true"></i>
         <span>{{'importData' | i18n}}</span>
     </button>
--- a/src/app/tools/import.component.html
+++ b/src/app/tools/import.component.html
@@ -1,7 +1,7 @@
 <div class="page-header">
     <h1>{{'importData' | i18n}}</h1>
 </div>
-<form #form (ngSubmit)="submit()" [appApiAction]="formPromise" ngNativeValidate>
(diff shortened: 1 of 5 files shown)


src/locales/en/messages.json
```diff
@@ -1000,6 +1000,12 @@
   "importData": {
     "message": "Import Data"
   },
+  "importError": {
+    "message": "Import Error"
+  },
+  "importErrorDesc": {
+    "message": "There was a problem with the data you tried to import. You can fix the errors below, or contact us if you need help."
```
cscharf, line 1007:
```suggestion
    "message": "There was a problem with the data you tried to import. Please resolve the errors listed below in your source file and try again."
```

src/app/tools/import.component.html
```diff
@@ -1,7 +1,7 @@
 <div class="page-header">
     <h1>{{'importData' | i18n}}</h1>
 </div>
-<form #form (ngSubmit)="submit()" [appApiAction]="formPromise" ngNativeValidate>
```
cscharf, line 4: does removing the `formPromise` default behavior of the `appApiAction` also remove the spinner/loading behavior as well? If so we'll need to replace that and ensure it's working.

cscharf (approved):

PR #860, opened 2021-03-02
Replace learn more link

vachan-maker:
Replaced the Learn More link of [PBKDF-2](https://en.wikipedia.org/wiki/PBKDF2) with https://bitwarden.com/help/article/what-encryption-is-used/#pbkdf2


cscharf (approved):
Nice, thanks @vachan-maker !
cc: @fschillingeriv

PR #862, opened 2021-03-02
changed send access creator identifier string

addisonbeck:
Clayton asked (& marketing agreed) to remove the 'A' from a Send's creator identifier blurb. 
I also moved that string to messages.json (whoops)


addisonbeck:
Not sure what that check failure is about

cscharf (approved):

PR #866, opened 2021-03-03
Expand encrypted export warning

eliykat:
## Objective
Add more information to the encrypted export warning, so that users know they can't import it into a different account.

This reflects our warning in the [help documentation](https://bitwarden.com/help/article/encrypted-export/).

## Code changes
Aside from the obvious changes in `messages.json`, I also tweaked our `platformUtils` `showDialog` method so that we have the option of using html markup in our modals. (required so that the warning wouldn't be a single blob of text)

Required changes in jslib: https://github.com/bitwarden/jslib/pull/294

### DIFF
--- a/src/locales/en/messages.json
+++ b/src/locales/en/messages.json
@@ -799,8 +799,11 @@
   "exportWarningDesc": {
     "message": "This export contains your vault data in an unencrypted format. You should not store or send the exported file over unsecure channels (such as email). Delete it immediately after you are done using it."
   },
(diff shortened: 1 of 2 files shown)


cscharf:
@eliykat , don't forget to do this on any other clients that support encrypted export. The jslib changes are in, however we'll need to get browser extension and desktop also updated (and mobile). Especially with the i18n key name changes.

cscharf (approved):

PR #877, opened 2021-03-08
Safari date time fix

addisonbeck:
Added `date-input-polyfill` to polyfills.ts to add date support in Safari, and added a select list of times for custom delete/expiration dates in Safari since `time` is not supported.


src/scss/plugins.scss
```diff
@@ -163,3 +163,9 @@ $fa-font-path: "~font-awesome/fonts";
         }
     }
 }
+
+date-input-polyfill {
+    &[data-open="true"] {
```
cscharf, line 168: Same formatting issue here as well.

cscharf (approved):

PR #880, opened 2021-03-09
Adds folder word-wrap

Leon-Hikari:
Uses spaces and dashes as preferred separator in folder names
(instead of just breaking wherever the max width is encountered)

**Current behavior:**
Folder names without word wrap

**New behavior:**
Folder names with word wrap

**Tested on:**
- Chrome
- Firefox
- Edge


eliykat (approved):
Thanks @Leon-Hikari, those screenshots show the change nicely. This seems like a simple UI improvement so I'm going to go ahead and merge it.

PR #916, opened 2021-04-07
[Reset Password] Rotate encryption key

vincentsalucci:
## Code Changes
- **change-password.component.ts**: Added daisy-chained method to rotate encryption key. For all confirmed orgs, re-enroll the `resetPasswordKey` if necessary.

### DIFF
--- a/src/app/settings/change-password.component.ts
+++ b/src/app/settings/change-password.component.ts
@@ -201,10 +201,12 @@ export class ChangePasswordComponent extends BaseChangePasswordComponent {
         const orgs = await this.userService.getAllOrganizations();
 
         for (const org of orgs) {
-            const request = new OrganizationUserResetPasswordEnrollmentRequest();
-            request.resetPasswordKey = null;
+            if (org.isResetPasswordEnrolled) {
+                const request = new OrganizationUserResetPasswordEnrollmentRequest();
+                request.resetPasswordKey = null;
 
-            await this.apiService.putOrganizationUserResetPasswordEnrollment(org.id, org.userId, request);
+                await this.apiService.putOrganizationUserResetPasswordEnrollment(org.id, org.userId, request);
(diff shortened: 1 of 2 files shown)


src/app/settings/change-password.component.ts
```diff
@@ -192,4 +196,15 @@ export class ChangePasswordComponent extends BaseChangePasswordComponent {
...
+    private async updateAllResetPasswordKeys() {
+        const orgs = await this.userService.getAllOrganizations();
+
+        for (const org of orgs) {
+            const request = new OrganizationUserResetPasswordEnrollmentRequest();
+            request.resetPasswordKey = null;
+
+            await this.apiService.putOrganizationUserResetPasswordEnrollment(org.id, org.userId, request);
+        }
+    }
```
kspearrin: This should be done on the server as part of the rotate operation, if possible. Having multiple actions like this on the client is more fail-prone.

Finally, should we consider updating the reset key for orgs rather than removing it?

vincentsalucci: The reason I put this here is because I believe our end goal will be to update the key rather than removing it (which will require client-side encryption). Phase 1 our of MVP is nulling the keys.

kspearrin: I'd argue that we should just go ahead and update it. It's not really any more effort here from what I can tell.

vincentsalucci: Fine by me! Will update task/docs accordingly.

vincentsalucci: @kspearrin - Updated. Please double check my encryption process. I want to make sure I used the correct `encKey` (gets a little confusing)

Removed warning string as we'll be taking care of each re-enrollment automagically.

kspearrin (approved):

PR #929, opened 2021-04-14
Password reprompt

Hinton:
## Objective
Adds support for requiring master password re-prompt before accessing hidden fields on ciphers. Also marks card numbers as a secret to ensure it's protected.

### Code Changes
- **src/app/vault/add-edit.component.html**: Change card number to be hidden. Add reprompt setting.
- **src/app/vault/bulk-actions.component.ts**: Ensure bulk actions asks for password if modifying protected item.
- **src/app/vault/ciphers.component.html**: Update values to be more consistent with other clients.
- **src/app/vault/ciphers.component.ts**: Protect viewing/copying values from ciphers list.
- **src/services/webPlatformUtils.service.ts**: Minor cleanup.

### Testing Considerations
We should verify all expected values are protected and cannot be accessed without prompting for master password.

### Screenshots
add edit
password prompt

Depends on https://github.com/bitwarden/jslib/pull/343


vincentsalucci (approved):

Virinum:
I have two questions on this feature:

1. Why is it not visible on https://vault.bitwarden.com/ ?
2. On my self hosted instance the checkbox isn't selected anymore after checking it, clicking "save" and reopen it. Also there is no re-prompt. Am I doing something wrong?

https://user-images.githubusercontent.com/42373861/120462268-0e2b5480-c39b-11eb-8d02-a1006e13bcf7.mp4

PR #935, opened 2021-04-19
[Reset Password] Admin Actions

vincentsalucci:
## Code Changes
- **jslib**: Update from 97ece68 -> 395ded0
- **accept-organization.component.ts**: Organized imports. Added `crypto.service` and `policy.service` to constructor. Added chained promise ability to auto-enroll a user into reset password. Uses policies retrieved via token to determine what actions to take.
- **login.component.html**: Added `app-callout` for organizations who are auto-enrolling users during the invite/acceptance flow.
- **login.component.ts**: Organized imports. Added `api.service` and `policy.service` to constructor.  Evaluated the org invite from storage to determine if the warning needed to be shown.
- **app-routing.module**: Updated `people` component to also check for the `ManageUsersPassword` permission before granting access to the component
- **app.module**: Added `ResetPasswordComponent` to the list of generated components/modals
- **people.component.html**: Added dropdown item for admins with proper permissions/params. Added modal template. Added visual indicator for org users who are enrolled in password reset.
- **people.component.ts**: Organized imports. Added `policy.service` to constructor. Added template reference. Added `canResetPassword` which designates the current user's permissions in regards to resetting a user's password. Added `orgUserResetPassword` which relates to the organization's plan ability. Added `orgHasKeys` which relates to the organization's public/private key state. Added `orgResetPasswordPolicyEnabled` which relates to the organization's current policy setup. Added `callingUserType` to evaluate calling user vs selected user. Created a couple of helpers for allowing reset password and when to show enrolled status. Added method to create/show/close the `ResetPasswordComponent` in a modal.
- **policies.component.ts**: Added reset password policy to the static listing.
- **policy-edit.component.html**: Added reset password policy UI. Added warning banners and data section.
- **policy-edit.component.ts**: Added data parsing for the reset password type
- **reset-password.component.html**: Created UI for changing a user's password. Includes a warning for what will happen to the user's sessions. Also includes a banner with all required master password policy.  Because we are enforcing all of the organization's master password policy parameters, the password strength meter was added as well as all of the basic validation checks for registering a user and creating a master password.
- **reset-password.component.ts**: Created component for hosting user password resets. Allows copying out the password to clipboard. Submitting will grab all necessary information for resetting the selected user's password (KDF, KDF iterations, `ResetPasswordKey`, organization private key). Then it will decrypt the organization's private key with the organization's share key, decrypt selected user's `ResetPasswordKey` with the decrypted private key, create a new key with the updated master password (and hash it), Encrypt the new key with the user's encryption key, and update the user table with the new hash and key.
- **account.component.ts**: Organized imports. Added `crypto.service` to constructor. Added logic to backfill organization pub/priv keys if necessary.
- **organization-type-guard.service**: Added `ManageUsersPassword` to the list of potential allowable permissions into the `people` component
- **change-password.component**: Updated conditional to use safer check for current user's enrollment status. TODO Update reset keys to use new flow!!
- **organization-plans.component.ts**: Organized imports. Added logic to create keys during new org creation. Also added backfill logic for pub/priv keys during organization upgrade.
- **organizations.component.html**: Updated conditional to use safer check for current user's enrollment status
- **organizations.component.ts**: Removed feature flag. Organized imports. Added `policy.service` to constructor. Added helper methods for allowing enrollment changes and showing the enrolled status per org. Updated encryption steps for enrolling.
- **messages.json**: Added necessary string changes

## Screenshots
0-enroll
1-enroll-warning
2-enroll-visual
3-withdraw
4-manage-permission
5-people-list
6-reset-modal
7-biz-portal-policies-list
8-biz-portal-edit
9-login-auto-enroll

## Dependency
- [Jslib admin actions](https://github.com/bitwarden/jslib/pull/349)
- [Jslib service updates](https://github.com/bitwarden/jslib/pull/387)


vincentsalucci:
@kspearrin - I don't have insight into the snyk test failing. Can you take a look (if necessary)

kspearrin:
Looks like we hit some kind of rate limit. It can be ignored.

src/app/organizations/manage/reset-password.component.html
```diff
@@ -0,0 +1,78 @@
...
+                    <small class="text-muted" *ngIf="name">{{name}}</small>
+                </h2>
+                <button type="button" class="close" data-dismiss="modal" appA11yTitle="{{'close' | i18n}}">
+                    <span aria-hidden="true">&times;</span>
+                </button>
+            </div>
+            <div class="modal-body">
+                <app-callout type="warning">{{'resetPasswordLoggedOutWarning' | i18n: loggedOutWarningName}}
+                </app-callout>
+                <app-callout type="info" *ngIf="enforcedPolicyOptions">
```
Hinton, line 16: It feels like we are duplicating this callout in many places, could we refactor it to be a component with `enforcedPolicyOptions` as input?

src/app/organizations/manage/reset-password.component.ts
```diff
@@ -0,0 +1,196 @@
...
+        let userInput: string[] = [];
+        const atPosition = this.email.indexOf('@');
+        if (atPosition > -1) {
+            userInput = userInput.concat(this.email.substr(0, atPosition).trim().toLowerCase().split(/[^A-Za-z0-9]/));
+        }
+        if (this.name != null && this.name !== '') {
+            userInput = userInput.concat(this.name.trim().toLowerCase().split(' '));
+        }
+        return userInput;
+    }
```
Hinton, line 195: Exists in 2 components (3 now), probably a good time to extract it shared code to ensure consistent behavior`.

src/app/organizations/manage/reset-password.component.ts
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        if (this.masterPasswordScore < 3) {
+            const result = await this.platformUtilsService.showDialog(this.i18nService.t('weakMasterPasswordDesc'),
+                this.i18nService.t('weakMasterPassword'), this.i18nService.t('yes'), this.i18nService.t('no'),
+                'warning');
+            if (!result) {
+                return false;
+            }
+        }
```
Hinton, line 123: Same here, would be nice if we could extract this and ensure we don't forget any rules in the future.

src/app/organizations/manage/reset-password.component.ts
```diff
@@ -0,0 +1,196 @@
...
+    async ngOnInit() {
+        // Get Enforced Policy Options
+        this.enforcedPolicyOptions = await this.policyService.getMasterPasswordPolicyOptions();
+    }
+
+    get loggedOutWarningName() {
+        return this.name != null ? this.name : this.i18nService.t('thisUser');
+    }
+
+    getPasswordScoreAlertDisplay() {
```
Hinton, line 53: Also duplicated.

src/app/accounts/accept-organization.component.ts
```diff
@@ -51,8 +60,39 @@ export class AcceptOrganizationComponent implements OnInit {
...
+                                    // RSA Encrypt user's encKey.key with organization public key
+                                    const encKey = await this.cryptoService.getEncKey();
+                                    const encryptedKey = await this.cryptoService.rsaEncrypt(encKey.key, publicKey.buffer);
+
+                                    // Create request and execute enrollment
+                                    const resetRequest = new OrganizationUserResetPasswordEnrollmentRequest();
+                                    resetRequest.resetPasswordKey = encryptedKey.encryptedString;
+
+                                    // Get User Id
+                                    // TODO Reset Password - Update this api method to use orgUser.Id
```
Hinton, line 86: TODO?

src/app/accounts/accept-organization.component.ts
```diff
@@ -51,8 +60,39 @@ export class AcceptOrganizationComponent implements OnInit {
...
+                                        throw new Error('Get Organization Keys response is null');
+                                    }
+
+                                    // RSA Encrypt user's encKey.key with organization public key
+                                    const encKey = await this.cryptoService.getEncKey();
+                                    const encryptedKey = await this.cryptoService.rsaEncrypt(encKey.key, publicKey.buffer);
+
+                                    // Create request and execute enrollment
+                                    const resetRequest = new OrganizationUserResetPasswordEnrollmentRequest();
+                                    resetRequest.resetPasswordKey = encryptedKey.encryptedString;
```
Hinton, line 83: These operations seems to be used in the org settings component as well, could we extract them to prevent duplicate code?

Hinton (commented):
I don't feel that I have sufficient context to give a "proper" review, but I've added a couple of opinionated comments. In general it feels like we are starting to duplicate a fair bit of code related to change password (I'm also guilty of doing this for emergency access) and should we ever add more checks it will be easy to miss places.

src/app/accounts/accept-organization.component.ts
vincentsalucci: @Hinton Does it make sense here to maybe make a new service at the `jslib` layer to house these re-used encryption steps?

Hinton: I don't believe we need to do these org steps in the other clients?

cscharf:
> I don't feel that I have sufficient context to give a "proper" review, but I've added a couple of opinionated comments. In general it feels like we are starting to duplicate a fair bit of code related to change password (I'm also guilty of doing this for emergency access) and should we ever add more checks it will be easy to miss places.

Agreed, @vincentsalucci , please put a card in Asana for some consolidation or cleanup, perhaps as we get to v1 after MVP of Admin Password Reset we can address some of these items and abstract a little better with a DRY approach.

src/app/organizations/manage/policy-edit.component.html
```diff
@@ -153,19 +156,32 @@ <h3 class="mt-4">{{'passphrase' | i18n}}</h3>
...
             </div>
             <div class="modal-footer">
                 <button type="submit" class="btn btn-primary btn-submit" [disabled]="form.loading">
                     <i class="fa fa-spinner fa-spin" title="{{'loading' | i18n}}" aria-hidden="true"></i>
                     <span>{{'save' | i18n}}</span>
                 </button>
-                <button type="button" class="btn btn-outline-secondary"
-                    data-dismiss="modal">{{'cancel' | i18n}}</button>
+                <button type="button" class="btn btn-outline-secondary" data-dismiss="modal">{{'cancel' |
+                    i18n}}</button>
```
cscharf, line 184: Odd place to wrap the button text binding

src/app/organizations/manage/reset-password.component.ts
```diff
@@ -0,0 +1,196 @@
...
+            if (!result) {
+                return false;
+            }
+        }
+
+        // Get user Information (kdf type, kdf iterations, resetPasswordKey, private key) and change password
+        try {
+            this.formPromise = this.apiService.getOrganizationUserResetPasswordDetails(this.organizationId, this.id)
+                .then(async response => {
+                    let kdfType = null;
```
cscharf, line 129: Small order of operations nit (but it's fine as-is, just perhaps in the future) when organizing logic, `const` handles memory better than `let` in the garbage collector although both handle context pretty well. Generally I would recommend testing the response first, if null throw an exception, THEN (outside of an `if/then` block) declare `const` and assign each of those variables. Makes things a little cleaner vs. assigning values which may never be used (declaring and assigning to `null` if we're only going to throw an exception anyway, etc.)

src/app/settings/organizations.component.ts
```diff
@@ -88,32 +104,56 @@ export class OrganizationsComponent implements OnInit {
...
+
+            // Retrieve Public Key
+            this.actionPromise = this.apiService.getOrganizationKeys(org.id)
+                .then(async response => {
+                    let publicKey = null;
+
+                    if (response != null) {
+                        publicKey = Utils.fromB64ToArray(response.publicKey);
+                    } else {
+                        throw new Error('Get Organization Keys response is null');
```
cscharf, line 129: Do need to localize error messages; although this one doesn't really mean anything to anyone other than perhaps Customer Success.

src/app/accounts/accept-organization.component.ts
vincentsalucci: Nope. Good point.

src/app/organizations/manage/policy-edit.component.html
vincentsalucci: Agreed - updated.

src/app/organizations/manage/reset-password.component.ts
vincentsalucci: Awesome bit of advice. Noted and updated.

src/app/settings/organizations.component.ts
vincentsalucci: Updated.

vincentsalucci:
> > I don't feel that I have sufficient context to give a "proper" review, but I've added a couple of opinionated comments. In general it feels like we are starting to duplicate a fair bit of code related to change password (I'm also guilty of doing this for emergency access) and should we ever add more checks it will be easy to miss places.
> 
> Agreed, @vincentsalucci , please put a card in Asana for some consolidation or cleanup, perhaps as we get to v1 after MVP of Admin Password Reset we can address some of these items and abstract a little better with a DRY approach.

Created Asana task for the above.

cscharf (approved):
Looks good, thanks Vince!

PR #937, opened 2021-04-19
New client configuration pattern

joseph-flinn:
## Summary
Creating a config pattern for the clients. As we add the new QA environment, it is important that we have a way to configure clients to easily point to the different environments (local dev, qa, and prod) as well as splitting out environment specific variables into these config files. 

There is a base config file (`config/base.json`) and the environment config files are used to override/add values. It defaults the defaults to local development vs production values.

## Files Changed
- **config.js:** Provides a function to load the correct config files
- **package.json:** Adding in a `build:qa` command
- **webpack.config.js:** Updating the webpack config script to use the new configuration files
- **config/base.json:** The base configuration file
- **config/qa.json:** the qa configuration file
- **config/production.json:** the production configuration file
- **README.md:** Updating the documentation to get a contributed started
- **.gitignore:** Ignoring the `config/development.json` file so local overrides aren't committed.

---
⚠️ Note: Secrets should **NOT** be stored in the config files. If there is a need to have a framework for storing secrets, we can implement them via a `.env` file and the `dotenv` library.


kspearrin:
Isn't this just for running locally via the webpack server? @Hinton  should probably review this.

Hinton:
@joseph-flinn I believe the only comment I have is that this will be a behavior change compared to how it currently works. Since it will point to the production instance by default instead of the local server. We need to update the readme to cover this as well as explain how to run against a local server.

config.js
```diff
@@ -0,0 +1,24 @@
+function load(envName) {
+    const supportedEnvironments = ['production', 'qa', 'development'];
+    
+    if (envName === undefined) {
+        envName = 'production'
+        console.log('config: No specific env config provided. Defaulting to production');
+    }
```
Hinton, line 7: We could consider moving the fallback env logic to webpack config since users might not immediately locate it currently.

joseph-flinn: I like it. I'll make that change

joseph-flinn:
> @joseph-flinn I believe the only comment I have is that this will be a behavior change compared to how it currently works. Since it will point to the production instance by default instead of the local server. We need to update the readme to cover this as well as explain how to run against a local server.

I don't see any advantage of production defaults over local defaults. It was just a pattern in a recent project that I was working on. I'll switch it to local defaults and update the documentation as well. If we find an advantage in the future, it isn't really that difficult to switch it back to production defaults.

config.js
joseph-flinn: Since we can no longer assume that `config/development.json` exists, this logic has changed a bit to allow for the case of no overrides.

I have followed the pattern to check for NODE_ENV being null in the webpack.config.js and also checking to see if the given value exists in the environment overrides that we provide. I have dropped any helpful user messages for those who are creating new environments that are not just local environments. But I did add a log of what is being passed in as the config to help in any debugging.

webpack.config.js
```diff
@@ -8,11 +8,22 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin');
...
 }
-const ENV = process.env.ENV = process.env.NODE_ENV;
+
+const NODE_ENV = process.env.NODE_ENV;
+const envConfig = config.load(process.env.ENV)
+config.log(envConfig)
+
+if (process.env.ENV == null) {
+    // Defaults to development if an ENV is not explicitly chosen. Also preserves the selfHost option
+    process.env.ENV = 'development'
```
Hinton, line 23: This seems to run after `config.load`. Seems reasonable to put it before?

package.json
```diff
@@ -13,8 +13,12 @@
     "symlink:lin": "rm -rf ./jslib && ln -s ../jslib ./jslib",
     "build": "gulp prebuild && webpack",
     "build:watch": "gulp prebuild && webpack-dev-server",
-    "build:prod": "gulp prebuild && cross-env NODE_ENV=production webpack",
-    "build:prod:watch": "gulp prebuild && cross-env NODE_ENV=production webpack-dev-server",
+    "build:dev": "gulp prebuild && cross-env ENV=development webpack",
```
Hinton, line 16: If default env is development we can remove the env from here.

README.md
```diff
@@ -43,23 +43,23 @@ npm install
 npm run build:prod:watch
 ```
 
-You can also manually adjusting your API endpoint settings in `webpack.config.js` by altering the `targets` within `proxy`. For example:
+You can also manually adjusting your API endpoint settings by adding `config/development.js` overriding any of the values in `config/base.json`. For example:
```
Hinton, line 46: We should mention that they can work directly against the production instance by setting the `ENV` environment variable to `production`. Since that will probably be the most common pattern for external contributors.

webpack.config.js
```diff
@@ -8,11 +8,22 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin');
...
+const NODE_ENV = process.env.NODE_ENV;
+const envConfig = config.load(process.env.ENV)
+config.log(envConfig)
+
+if (process.env.ENV == null) {
+    // Defaults to development if an ENV is not explicitly chosen. Also preserves the selfHost option
+    process.env.ENV = 'development' 
+} 
+
+const ENV = process.env.ENV;
```
Hinton, line 26: It this used anywhere?

joseph-flinn: It is used once on line 137. I'm not sure exactly what that `DefinePlugin` is used for but it looked like that was what the rest of the code was going to use so I wanted to expose the ENV environments vs just the NODE_ENV environments.

webpack.config.js
joseph-flinn: After chatting, I've changed the logic to assume that there is not an initial `development.json` override file. The default values (stored in `base.json`) happen to be the development values. I was thinking to provide a default experience without requiring the setup step of creating `development.json` previous to being able to run the project, but allowing the flexibility to override the default experience with the `development.json`

This placement allows us to preserve the choice (or the lack of) while running webpack to load the config files

### Current cases:
#### [ `base.json` ], `cross-env NODE_ENV=development webpack`
all default values that we provide. This is the default if no ENV is passed through. Most likely, `development.json` doesn't exist in this case.

#### [ `base.json`, `development.json` ], `cross-env NODE_ENV=development ENV=development webpack`
override some/all default dev values provided by `base` with what is in the user created `development.json`

#### [ `base.json`, `qa.json` ], `cross-env NODE_ENV=production ENV=qa webpack`
overrides the default dev values to run against the QA environment. 

#### [ `base.json`, `prod.json` ], `cross-env NODE_ENV=production ENV=production webpack`
overrides the default values to run against the production environment.

webpack.config.js
Hinton: I believe it should actually be the NODE_ENV environment, since it's now used to determine log level and enable angular production mode.

Hinton: Which should let you remove line 21-26.

README.md
```diff
@@ -43,23 +43,29 @@ npm install
 npm run build:prod:watch
```
Hinton, line 43: Now that I re-read the readme I believe we should switch out this instruction with the following since it will run everything in development mode which provides better stack traces and logging.

```suggestion
ENV=production npm run build:watch ´
```

README.md
```diff
@@ -43,23 +43,29 @@ npm install
...
 ```
 
-You can also manually adjusting your API endpoint settings in `webpack.config.js` by altering the `targets` within `proxy`. For example:
+Or you can set the ENV variable to `production` and run webpack manually.
+
+```
+cross-env NODE_ENV=production ENV=production webpack
+```
+
```
Hinton, line 51: And this can probably be removed since I missed that we have more npm run commands.
```suggestion
```

webpack.config.js
joseph-flinn: Perfect! I'll get that switched.

Hinton (approved):
Looks good!

PR #953, opened 2021-04-27
Add auto delete warning to trash page

Hinton:
## Objective
Now that trashed items will be automatically deleted after a while we should notify users. Since we don't know the frequency for self-hosted instances there are two different phrasings one specifying 30 days and one generic.

### Screenshots
image

### DIFF
--- a/src/app/organizations/vault/vault.component.html
+++ b/src/app/organizations/vault/vault.component.html
@@ -29,9 +29,9 @@
                     </button>
                 </div>
             </div>
-            <div class="alert alert-warning" *ngIf="deleted" >
+            <app-callout type="warning" *ngIf="deleted" icon="fa-warning">
                 {{trashCleanupWarning}}
-            </div>
+            </app-callout>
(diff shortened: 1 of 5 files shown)


src/app/vault/vault.component.html
```diff
@@ -28,6 +28,9 @@ <h1>
                     </button>
                 </div>
             </div>
+            <div class="alert alert-warning" *ngIf="deleted" >
```
kspearrin, line 31: use the `app-callout` component to keep styling in line with other places around the web vault

src/app/organizations/vault/vault.component.html
```diff
@@ -29,6 +29,9 @@ <h1>
                     </button>
                 </div>
             </div>
+            <div class="alert alert-warning" *ngIf="deleted" >
```
kspearrin, line 32: use the `app-callout` component to keep styling in line with other places around the web vault

kspearrin (changes requested):

kspearrin (approved):

PR #967, opened 2021-05-11
Add event export

MGibson1:
# Overview

Closes https://app.asana.com/0/1199950673439194/1155011472108808/f
Related to bitwarden/jslib#375

Adds the ability to export Organization events to a csv.

# Files Changed
* **entity-events.component.ts**: update eventService call to async
* **event.component.html**: Add export button.
* **events.component.ts**: Add events export and update eventService get event info to async
* **event.service.ts**: Add human readable information to event info. This is to support export since we don't want html in our export CSVs. All things we use html for can be searched for by the `shortId` except Policy. Here we need to convert the policyId to a policy type for the organization. This is done through an async call to policyService, which bubbles the whole thing async up to the component level.

### Requires bitwarden/jslib#375

### DIFF
--- a/src/app/organizations/manage/events.component.ts
+++ b/src/app/organizations/manage/events.component.ts
@@ -14,6 +14,7 @@ import { EventService } from '../../services/event.service';
 
 import { EventResponse } from 'jslib/models/response/eventResponse';
 import { ListResponse } from 'jslib/models/response/listResponse';
+import { EventView } from 'jslib/models/view/eventView';
 
 @Component({
     selector: 'app-org-events',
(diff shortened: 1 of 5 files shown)


src/app/organizations/manage/events.component.ts
```diff
@@ -64,8 +68,25 @@ export class EventsComponent implements OnInit {
...
+            return;
+        }
+
+        this.loading = true;
+        try {
+            this.exportPromise = this.exportService.getEventExport(this.events);
+            const data = await this.exportPromise;
+            const fileName = this.exportService.getFileName('org-events', 'csv');
+            this.platformUtilsService.saveFile(window, data, { type: 'text/plain' }, fileName);
+        } catch { }
```
eliykat, line 82: I thought we were moving away from empty `try/catch` blocks (or at least putting a debug message in there). Is there any particular error you're expecting here that you want to swallow?

src/app/organizations/manage/events.component.ts
```diff
@@ -120,4 +142,11 @@ export class EventsComponent implements OnInit {
         this.morePromise = null;
         this.refreshPromise = null;
     }
+
+    private appApiPromiseUnfulfilled() {
+        if (this.refreshPromise != null || this.morePromise != null || this.exportPromise != null) {
```
eliykat, line 147: Couldn't you just return this value directly instead of using the `if` block? Unless this is for readability or following a pattern

eliykat (commented):
This all made sense to me (including the related jslib changes), just a couple of small style-related questions. However I haven't touched Events until now so it's worth someone else running their eyes over as well.

src/app/organizations/manage/events.component.ts
MGibson1: nice, yep :)

src/app/organizations/manage/events.component.ts
MGibson1: I totally agree. I didn't because the error handling is done in [api-action.directive](https://github.com/bitwarden/jslib/blob/master/src/angular/directives/api-action.directive.ts)

Now that I look again, though I want to combine all the save stuff in there too.

eliykat (approved):
Yeah alright, go on then

PR #969, opened 2021-05-12
Change WebAuthn connectors to not use inline onclick

Hinton:
## Objective
The cloud web vault prevents inline js using CSP rules. Which breaks the WebAuthn buttons. To work-around this I set the `onclick` handler in the script files instead.

### Code Changes
- **src/connectors/webauthn(-fallback).html**:  Remove inline `onclick`.
- **src/connectors/webauthn(-fallback).ts**: Set `onclick` handler.

### DIFF
--- a/src/connectors/webauthn-fallback.ts
+++ b/src/connectors/webauthn-fallback.ts
@@ -19,7 +19,10 @@ document.addEventListener('DOMContentLoaded', async () => {
 
     document.getElementById('msg').innerText = translate('webAuthnFallbackMsg');
     document.getElementById('remember-label').innerText = translate('rememberMe');
-    document.getElementById('webauthn-button').innerText = translate('webAuthnAuthenticate');
+
+    const button = document.getElementById('webauthn-button');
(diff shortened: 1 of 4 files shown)


addisonbeck (approved):

src/connectors/webauthn-fallback.ts
```diff
@@ -19,7 +19,10 @@ document.addEventListener('DOMContentLoaded', async () => {
 
     document.getElementById('msg').innerText = translate('webAuthnFallbackMsg');
     document.getElementById('remember-label').innerText = translate('rememberMe');
-    document.getElementById('webauthn-button').innerText = translate('webAuthnAuthenticate');
```
cscharf, line 22: Why remove the button text?

cscharf (changes requested):

src/connectors/webauthn-fallback.ts
Hinton: It's set on line 24. I extracted the button to a variable.

cscharf (approved):

PR #976, opened 2021-05-14
Change all remaining modals to be scrollable

Hinton:
## Objective
In #957 I changed the add-cipher modal to be scrollable. After some discussion we concluded that several other modals would also benefit from being scrollable. This PR changes all remaining modals to be scrollable, except for two-factor settings since the two-factor-validator interferes.

### Testing Considerations
We should do some regression testing on all modals and verify they still behave as expected.

### DIFF
--- a/src/app/accounts/two-factor-options.component.html
+++ b/src/app/accounts/two-factor-options.component.html
@@ -7,17 +7,19 @@
                     <span aria-hidden="true">&times;</span>
                 </button>
             </div>
-            <div class="list-group list-group-flush">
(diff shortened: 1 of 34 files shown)


addisonbeck (dismissed):

Hinton:
Had to revert the modal-dialog-scrollable on two-factor-components since it breaks if there is any non-expected html tag between modal and modal-content. I tried a few different approaches but it seems we would need to redo how the verify components works to solve it properly.

addisonbeck (approved):

PR #981, opened 2021-05-19
Increase 'MaxCheckedCount' 

colinkr:
To prevent Bug, when selecting more than 500 Entries.


cscharf:
Hi @colinkr , thank you for the submission, however at this time we're keeping the 500 item selection limit as currently implemented. Also, increasing this limit to 1000 simply "moves the issue" upstream without actually providing any solutions or usability workarounds which would not be adequate.

PR #986, opened 2021-05-20
Fix accessibility (a11y) on swal2 modals

### LINKED ISSUE #948: Accessibility - Cannot TAB to 'Delete Folder' confirmation window and NVDA cannot reach and read it.
## Describe the Bug
When deleting a folder from the My Vault page, the 'Delete Folder' confirmation window is not in focus and you cannot TAB to it, and the NVDA screen reader cannot read the message or the buttons.

## Steps To Reproduce
1. Log into your web vault.
2. Create a test folder from the My Vault view.
3. Try to delete the test folder
4. Try to TAB to reach the buttons or message so the screen reader would read them. 

## Expected Result
To be able to reach the new window using TAB.

## Actual Result
You cannot TAB to the new 'Delete Folder' confirmation window.

## Screenshots or Videos
N/A.

## Environment
- Operating system: Windows 10 Pro.
- Browser: Firefox 88.0.
- Build Version: 2.19.0.
- NVDA: 2020.4.

## Additional Context
After opening the 'Delete Folder' confirmation window, pressing the Esc key closes the 'EDIT FOLDER' window in the background.

eliykat:
## Objective

Fix #948. Keyboard accessibility of swal2 modals is broken when they pop up over the top of bootstrap modals. For example:
* if you edit a folder (opens a bootstrap modal) and you click delete (opens a swal2 modal to confirm)
* if you edit a login item (bootstrap modal) and you try to regenerate the password (opens a swal2 modal to warn about overwriting existing password)

In these situations, the boostrap modal still has keyboard focus, so:
* screenreaders will not read the swal2 modal
* tab will only cycle focus within the bootstrap modal
* hitting the esc key will close out of the boostrap modal and leave the swal2 modal open

## Code changes

* `webPlatformUtils`: before opening a swal2 dialog, remove `tabindex="-1"` from the bootstrap modal (if any). This lets the swal2 modal take focus. Restore `tabindex="-1"` after the swal2 modal is closed so that bootstrap can recapture focus.

I tested this in FF, Safari and Chrome on MacOS; and with NVDA on FF on Windows, and it works consistently.

### DIFF
--- a/src/services/webPlatformUtils.service.ts
+++ b/src/services/webPlatformUtils.service.ts
@@ -195,9 +195,9 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
             }
         }
 
-        const bootstrapModal = document.querySelector("div.modal");
+        const bootstrapModal = document.querySelector('div.modal');
         if (bootstrapModal != null) {
-            bootstrapModal.removeAttribute("tabindex");
+            bootstrapModal.removeAttribute('tabindex');
         }
 
         const iconHtmlStr = iconClasses != null ? `<i class="swal-custom-icon fa ${iconClasses}"></i>` : undefined;
@@ -216,7 +216,7 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
         });
 
         if (bootstrapModal != null) {
-            bootstrapModal.setAttribute("tabindex", "-1");
+            bootstrapModal.setAttribute('tabindex', '-1');
         }
 
         return confirmed.value;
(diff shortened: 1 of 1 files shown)


addisonbeck (approved):
Just curious, did you try setting `tabindex` of the swal model to a deeper negative value, like `-2`? The tabindex [spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) doesn't touch on this much, but it does look like you _can_ go further with negative numbers, though I'm not sure if it'll have an effect.

If that isn't an option, this seems okay to me.

eliykat:
I just tried _going deeper_ on the `tabindex` but it didn't change anything, so I'll just merge this.

PR #994, opened 2021-05-26
Set baseUrl in QA cloud environment

eliykat:
## Objective
As discussed with @cscharf and @joseph-flinn, our QA cloud environment does not have its environment URLs set properly. e.g. Send links still point to the prod servers.

This change will set the QA environment's `baseURL` to `window.location.origin`, which is the same way self-hosted instances work.

We agreed that it would be best to have fully configurable environment URLs set at runtime (as with other Angular clients), but  it's out of scope of the current fix.

## Code changes
`services.module.ts` - per @cscharf's suggestion, change the test for when we explicitly set the `baseURL`. Previously it tested `isDev() || platformUtilService.isSelfHost()`, but QA is neither of these things. Really we just want to test whether it's running on the prod servers or not. We discussed passing environment variables etc, but I thought that the most straightforward solution is just to refer to `window.location.origin`. Really we just want to test the _location_ of the deployment, and passing more environment variables to do this seemed unnecessary.

Because this is only tested in this one place, I have not lifted it up to something like `platformUtilsService.isDeployedToProd()`.

I also did a global search for `bitwarden.com` to make sure that this would filter through to everywhere required. I'm fairly sure it will, as everywhere seems to refer to `environmentService.getWebVaultUrl()`. That said, it won't update any external links/redirects (e.g. to help docs or to account management), but I assume they are not a priority and/or not a problem.


joseph-flinn:
I built and deployed this to the QA environment and it works as expected. The send and notification urls work out of the box with a change needed to the QA proxy to pick up the correct icons url

src/app/services/services.module.ts
```diff
@@ -144,9 +144,8 @@ containerService.attachToWindow(window);
 export function initFactory(): Function {
     return async () => {
         await (storageService as HtmlStorageService).init();
-        const isDev = platformUtilsService.isDev();
 
-        if (isDev || platformUtilsService.isSelfHost()) {
+        if (window.location.origin.indexOf('vault.bitwarden.com') === -1) {
```
cscharf, line 148: This needs to be protected either using the environment variable approach OR to account for malicious (or well-meaning but poorly selected) sub-domain hijacking/spoofing (by using an explicit authority test): e.g., `https://vault.bitwarden.com.badactor.compromised-site.tv`, which would still pass this test and run with production URLs in the web vault for notifications and the business portal.

eliykat: Ah good point! back to environment variables then

eliykat:
I've taken a different approach, using environment variables as recommended, but I reworked a few things in the process.

* we had an environment variable, `ENV`, which was set to `qa`, `development` or `production` depending on the deployment target.
* however, `ENV` was only used to load the webpack config. In the runtime environment, `process.env.ENV` was set to the `NODE_ENV` environment variable instead. This was super confusing and meant that we didn't have access to `ENV`.
* so what I've done is:
  * rename `ENV` to `TARGET` to reduce confusion with `NODE_ENV`
  * make `TARGET` available in the runtime environment, and test it in `platformUtilsService.isProdServer()`. This is then used to set the baseUrl.
  * change the `SELF_HOST` variable to be a value of `TARGET` instead, and update the test in `platformUtilsService.isSelfHost()`. This is optional but it felt like a simplification that made sense.

joseph-flinn (changes requested):
I think this switch in approach is a great idea. I just came across another place where the `isDev` logic is used where I'll need to pull the target environment value.

I have one concern: right now as it stands, I could set ENV to `development` through NODE_ENV and set TARGET to `production`. `isDev()` would return `true` while `process.env.ENV` is `production`. I'm wondering if we can reinstate the SELFHOST env var and pass the TARGET/ENV through to ENV versus the NODD_ENV (in the webpack config)

cscharf (approved):
Looks good, thank you @eliykat for your help!

joseph-flinn (approved):

eliykat:
Thanks for finishing this one off! However, correct me if I'm missing something, but self-hosted also runs in `prod` mode, so won't the [new test in services.module.ts](https://github.com/bitwarden/web/blob/7e2cc661be20bea7e101f188caae57ee77e4dfa0/src/app/services/services.module.ts#L148) now set official Bitwarden URLs for self-hosted servers as well?

My thinking was that `TARGET` referred to the location of the deployment, and `ENV` referred to the mode it was running in. You can have prod mode running on prod servers, prod mode running on self-hosted servers, even prod mode running on localhost if you were testing something.

That said, if we want to keep it in 1 variable, maybe we just need to check for self-hosted as well in `services.module.ts`?

joseph-flinn:
Ah, good catch. I forgot to add the SelfHost check back in.

PR #997, opened 2021-05-27
adding the self host check back in

joseph-flinn:
## Summary
As discovered and discussed with @eliykat, in the previous change, the selfhost check was mistakenly taken out of the `services.modules.ts`. This adds the check back in.


eliykat (approved):

joseph-flinn:
@eliykat Thanks for the catch!

PR #999, opened 2021-05-27
Additional timeouts for the full working day

dltmurphy:
In relation to the [Longer Lock Time](https://community.bitwarden.com/t/custom-and-longer-lock-time/77/22) requests that are most recent on the forum

I tested it by setting it to 30 seconds and timing it to ensure it was accurate and I hadn't missed any other configuration

I'm looking into the custom option people have been requesting but I feel it's worthwhile merging this in the interim as it will take me a few weeks to get the custom one working properly 😊

### DIFF
--- a/src/app/settings/options.component.ts
+++ b/src/app/settings/options.component.ts
@@ -43,6 +43,8 @@ export class OptionsComponent implements OnInit {
             { name: i18nService.t('thirtyMinutes'), value: 30 },
             { name: i18nService.t('oneHour'), value: 60 },
             { name: i18nService.t('fourHours'), value: 240 },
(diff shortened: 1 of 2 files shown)


eliykat:
Hi @dltmurphy, thanks for the PR. As outlined [here](https://community.bitwarden.com/t/add-8-12-and-24-hours-timeout-options/29996/7?u=eliykat), we're not looking at adding any additional fixed timeout options (even as an interim solution). Instead, any change to this feature should allow for a custom timeout. If you'd like to help, I would encourage you to coordinate on the [existing PR](https://github.com/bitwarden/browser/pull/1851) to avoid any duplication of effort.

PR #1010, opened 2021-06-02
Add deploy workflow

joseph-flinn:
## Summary
With one of the last Web client releases, there was an outage when the process stopped halfway through the build. Automating the process will help prevent outages like this in the future.

The new deploy workflow can be triggered in two ways: 1) draft release being published, or 2) manual trigger with release tag version. The publish trigger works like most of the other clients. As soon as you publish the release with the release notes, that pipeline is triggered to deploy the site. For some reason, if the deploy workflow fails, the manual trigger is there as a back up so that we can deploy an already released version (more like the current contents of `rc`). 

## Notes
- The release/deploy flow has been fully tested on a personal repository

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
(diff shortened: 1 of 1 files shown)


joseph-flinn:
@cscharf Tagging for visibility

vgrassia (approved):

PR #974, opened 2021-05-13
Theme Support with a Dark Mode

dltmurphy:
This is in reference to the Dark Mode feature requests below:

[Dark Mode Feature Suggestion](https://community.bitwarden.com/t/dark-mode-for-web-vault/6790)

[Dark Mode Contribution Forum](https://community.bitwarden.com/t/web-vault-dark-mode/29756)

The options section successfully changes the theme when you select the other theme, however it doesn't save it to the account so after logging out and back in or refreshing the page, you're back to the default used by the browser

Theming in variables.scss should allow easy addition for Nord and Solarised themes that are present in the extension already

Remade from [958](https://github.com/bitwarden/web/pull/958) after misconfiguration of git

PR #1017, opened 2021-06-07
Dark Theme

dltmurphy:
Issues from [974](https://github.com/bitwarden/web/pull/974) should now be resolved but please test and let me know any issues


dltmurphy:
What's the preference on the variables file in terms of the naming? Shall i update them to be listed as colours rather than descriptions (and update it in the other repos) so that the file is smaller and easier to maintain and read or is it better to be the descriptions?

For example from this:
```
textColor: #fbfbfb,
textMuted: #C1C4C8,
linkColor: #52bdfb,
linkColorHover:#2ba0e3,
iconHover: #555555,
iconColor: #6c757d,
```

To this:
```
white1: #fbfbfb,
grey1: #C1C4C8,
blue1: #52bdfb,
blue2: #2ba0e3,
grey2: #555555,
grey3: #6c757d,
```

cscharf:
@theegreatsamboni , in order to test/review locally, please use the GitHub command-line tool, `gh pr checkout 1017` from your local web repo and you should be able to pull and run this one. Let me know if you run into any issues. @fschillingeriv , same deal, if you could, but perhaps this is something where Sam could share some screen shots with you based on your prior feedback.

src/services/htmlStorage.service.ts
```diff
@@ -26,6 +26,12 @@ export class HtmlStorageService implements StorageService {
         if (vaultTimeoutAction == null) {
             await this.save(ConstantsService.vaultTimeoutActionKey, 'lock');
         }
+
+        // Default theme to match the browser if the theme isn't set
+        const theme = await this.get<string>(ConstantsService.themeKey);
+        if (theme == null) {
```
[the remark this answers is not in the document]
dario-piotrowicz: Actually I can see the double equals a bit all over the place here... I don't really get why :confused: 
do you know?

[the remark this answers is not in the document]
dltmurphy: Its catching undefined too and it didn't work when i tried it with three without setting the variable in the options. 
I've tried it with `if(!theme)` which works. Do you want me to commit this (and also the `if (obj instanceof Set)` that is missing after my rebase)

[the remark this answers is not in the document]
dario-piotrowicz: yeah of course I meant `if(!theme)` and not `if(theme)` such a blunder! sorry :P

Anyways no please keep it as `if (theme == null) {` as that seems the common pattern here, I was just curious if you had any insights as to why that is :)

About the `if ( obj instanceof Set )` yeah that does seem unrelated to your changes and I think you should definitely put it back :)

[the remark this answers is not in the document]
dario-piotrowicz: Thanks a bunch @eliykat I just asked as I am not that used to seeing that :)

src/app/settings/options.component.ts
```diff
@@ -101,4 +110,22 @@ export class OptionsComponent implements OnInit {
         }
         this.vaultTimeoutAction = newValue;
     }
+
+    async themeChanged(themeUpdate: string) {
```
eliykat, line 114: I think that the theme should not update until the user hits the "Save" button. I think the idea of previewing it without actually saving the changes it is confusing. It's also inconsistent with other options, which generally do not take effect until the user clicks the Save button. @cscharf might disagree though.

src/services/htmlStorage.service.ts
```diff
@@ -26,6 +26,12 @@ export class HtmlStorageService implements StorageService {
...
+            await this.save(ConstantsService.themeKey, 'theme_defaultSet');
```
eliykat, line 33: This should also not be necessary once you make default theme == null. It'll be null by default.

eliykat (changes requested):
Thanks for this @dltmurphy. I've only looked at the code so far and left some comments for you, I'll look at the aesthetic side of things next week.

Our desktop app responds to theme changes on the fly, i.e. without reloading the window. It looks like you could do the same here by attaching an event listener to the media query ([I found this example](https://web.dev/prefers-color-scheme/#reacting-on-dark-mode-changes)). This is not required for merging, but you seem keen so I thought I'd mention it.

If you decide to implement this, here's how to do it in a way that's consistent with our codebase:
* `platformUtilsService.onDefaultSystemThemeChange` is not currently implemented for the web client. Update it to a callback function as an argument, and pass that callback to the event listener for whatever the theme change event is. ([example](https://github.com/bitwarden/jslib/blob/master/electron/src/services/electronPlatformUtils.service.ts#L204) from desktop client)
* `services.module.ts` calls the above method and supplies the callback function that changes the html classes to change the theme ([example](https://github.com/bitwarden/desktop/blob/master/src/app/services.module.ts#L159) from desktop). However, this should only be called if theme is default.

eliykat:
> What's the preference on the variables file in terms of the naming? Shall i update them to be listed as colours rather than descriptions (and update it in the other repos) so that the file is smaller and easier to maintain and read or is it better to be the descriptions?

You should keep the same code style as other repos, which I think is what you've done currently, so no need to change this.

dltmurphy:
@eliykat 
> is there any way to stop the loading screen flashing up white/bright before the dark theme loads in?

I've added in media queries which will match the browser's theme for the initial index page. The theme doesn't kick in until after this point so not sure how much we can do to get this to style based on their options

> Our desktop app responds to theme changes on the fly

I'd like to look at that in another PR once this one is merged in. I have a few changes to make on my local dev to be able to facilitate it better

I'm running through the comments now on the logic of the options and will reply to them shortly

src/app/settings/options.component.html
```diff
@@ -87,6 +87,17 @@ <h1>{{'options' | i18n}}</h1>
         </div>
         <small class="form-text text-muted">{{'enableFullWidthDesc' | i18n}}</small>
     </div>
+    <div class="row">
```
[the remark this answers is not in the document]
dltmurphy: Sorted the indent. Turned out to just be line 80 making the rest look skewed

The select inputs are all wrapped in a col-6 and a row. Without these the select is the full width of the form instead:
image

dltmurphy:
Hope these changes have addressed everything @eliykat, let me know if i've missed anything or done anything illogical :-)

src/app/settings/options.component.ts
dltmurphy: We could get it to save the setting whenever this particular option is updated like in the browser, what do you think of doing this? It's beneficial to preview the theme first but i agree its confusing as you can click away from the settings whilst previewing the theme and easily forget to save it

eliykat: I think just apply and save the theme when the user clicks the Save button. I know other clients don't work this way (e.g. browser) but I think it's more important to be consistent with the other options on this page of this client.

src/services/htmlStorage.service.ts
```diff
@@ -5,7 +5,7 @@ import { ConstantsService } from 'jslib-common/services';
 export class HtmlStorageService implements StorageService {
     private localStorageKeys = new Set(['appId', 'anonymousAppId', 'rememberedEmail', 'passwordGenerationOptions',
         ConstantsService.disableFaviconKey, 'rememberEmail', 'enableGravatars', 'enableFullWidth',
-        ConstantsService.localeKey, ConstantsService.autoConfirmFingerprints,
+        ConstantsService.themeKey, ConstantsService.localeKey, ConstantsService.autoConfirmFingerprints,
```
eliykat, line 8: I think you can revert this change now that you're not making any other changes in this file.

src/services/webPlatformUtils.service.ts
```diff
@@ -313,11 +314,13 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
...
-    getDefaultSystemTheme() {
-        return Promise.resolve(null as 'light' | 'dark');
+    getDefaultSystemTheme(): Promise<'light' | 'dark'> {
+        return Promise.resolve(this.prefersColorSchemeDark.matches ? 'dark' : 'light');
     }
 
-    onDefaultSystemThemeChange() {
-        /* noop */
+    onDefaultSystemThemeChange(callback: ((theme: 'light' | 'dark') => unknown)) {
+        this.prefersColorSchemeDark.addListener(({ matches }) => {
```
eliykat, line 322: I don't know much about media queries, but `addListener` appears to be deprecated: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener. Is there a more future-proof way you can do this?

src/app/settings/options.component.html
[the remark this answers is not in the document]
eliykat: My bad, I didn't spot that the `<form-group>` tag is below - indentation was throwing me off.

src/services/webPlatformUtils.service.ts
dltmurphy: Sorted, I'll do a PR for fixing this in the browser too

dltmurphy: PR for the browser: [#1898](https://github.com/bitwarden/browser/pull/1898)

eliykat:
Hi @dltmurphy, I've asked our design team to review the visual design of this dark mode. We need to make sure the colours look good, meet our brand requirements, and meet accessibility standards. They're the experts in this area so I'll wait to hear back from them before we progress it here. I'll let you know any feedback they have.

eliykat:
@dltmurphy, can you please give me your email address so that I can share this design doc with you? If you want to keep it private, you can DM me on [Gitter](https://gitter.im/bitwarden/Lobby).

dltmurphy:
@eliykat - okay I've messaged you it on Gitter 😊

src/app/settings/options.component.ts
```diff
@@ -26,11 +26,15 @@ export class OptionsComponent implements OnInit {
...
     enableGravatars: boolean;
     enableFullWidth: boolean;
+    theme: string = null;
     locale: string;
     vaultTimeouts: any[];
     localeOptions: any[];
+    themeOptions: any[];
 
     private startingLocale: string;
+    private startingTheme: string;
```
eliykat, line 36: I think `currentTheme` would be a clearer name

src/app/settings/options.component.ts
```diff
@@ -80,6 +90,16 @@ export class OptionsComponent implements OnInit {
         await this.stateService.save('enableGravatars', this.enableGravatars);
         await this.storageService.save('enableFullWidth', this.enableFullWidth);
         this.messagingService.send('setFullWidth');
+        if (this.theme !== this.startingTheme) {
+            this.changeTheme = this.startingTheme = this.theme;
```
eliykat, line 94: Unless I'm overlooking something, `changeTheme` doesn't need to be a property - it can just be a local variable declared with `let`. 

Also, maybe `effectiveTheme` would be a clearer name? (It's kind of difficult to distinguish between what you've chosen vs. what actual theme to apply)

src/app/settings/options.component.ts
```diff
@@ -80,6 +90,16 @@ export class OptionsComponent implements OnInit {
...
+            await this.storageService.save('theme', this.changeTheme);
```
eliykat, line 95: Use `this.theme` as the source to save - it's all the same value at the moment, but it makes it clearer that you're saving the user's selection, and not the "effective" theme.

src/app/settings/options.component.ts
```diff
@@ -26,11 +26,15 @@ export class OptionsComponent implements OnInit {
...
     enableFullWidth: boolean;
+    theme: string = null;
     locale: string;
     vaultTimeouts: any[];
     localeOptions: any[];
+    themeOptions: any[];
 
     private startingLocale: string;
+    private startingTheme: string;
+    private changeTheme: string;
```
eliykat, line 37: Per my comment below, I don't think this needs to be a class property

src/locales/en/messages.json
```diff
@@ -3988,6 +3988,24 @@
   "removeUsersWarning": {
     "message": "Are you sure you want to remove the following users? The process may take a few seconds to complete and cannot be interrupted or canceled."
   },
+  "usersHasBeenRemoved": {
```
eliykat, line 3991: What's this about?

eliykat (changes requested):
I've requested a few changes on the theme change logic, but overall it seems like the right approach to me. There's also a random UI string that got pulled in somewhere along the way that I think you'll have to undo.

I'll wait until the visual design is settled before I look at the other changes - we'll chat more about that separately.

src/locales/en/messages.json
dltmurphy: This is left over from a rebase i believe. I'll remove it

src/app/settings/payment.component.ts
```diff
@@ -87,6 +81,38 @@ export class PaymentComponent implements OnInit {
...
+            this.theme = await this.platformUtilsService.getDefaultSystemTheme();
+        }
+        if (this.theme === 'light') {
+            this.color = lightInputColor;
+            this.placeholder = lightInputPlaceholderColor;
+        } else {
+            this.color = darkInputColor;
+            this.placeholder = darkInputPlaceholderColor;
+        }
+        this.StripeElementStyle = {
```
eliykat, line 95: I'm OK with this being a class property now that we're changing it during runtime. However, I'd structure it as follows:
* `this.StripeElementStyle` and `this.StripeElementClasses` - move these initial assignments to the constructor. Set the `color` values to `null` in the constructor, so that they can be initialized without needing to access to `storageService`.
* then when you're setting the theme colours in `ngOnInit`, instead of having `this.color` and `this.placeholder`, you can just set the values directly, e.g. `this.StripeElementStyle.base.color = lightInputColor`. (or whatever colour it needs to be)
* this also means you can remove `this.color` and `this.placeholder`.

Different way of achieving the same thing, however it feels more straightforward without having so many variables and properties floating around when they're only used in one location.

src/app/settings/payment.component.ts
```diff
@@ -8,27 +8,16 @@ import { PaymentMethodType } from 'jslib-common/enums/paymentMethodType';
...
-        color: '#333333',
-    },
-};
-
-const StripeElementClasses = {
-    focus: 'is-focused',
-    empty: 'is-empty',
-    invalid: 'is-invalid',
-};
+const lightInputColor = '#465057';
```
eliykat, line 16: This is much better 👍

eliykat (changes requested):
As discussed on Gitter, here are my thoughts on structuring the Stripe panel.

src/app/settings/payment.component.ts
eliykat: I've pushed this change as discussed

src/app/settings/payment.component.ts
```diff
@@ -8,27 +8,15 @@ import { PaymentMethodType } from 'jslib-common/enums/paymentMethodType';
 
 import { ApiService } from 'jslib-common/abstractions/api.service';
 import { PlatformUtilsService } from 'jslib-common/abstractions/platformUtils.service';
+import { StorageService } from 'jslib-common/abstractions/storage.service';
+import { ConstantsService } from 'jslib-common/services/constants.service';
```
eliykat, line 12: This should be in its own import group (i.e. separated by a newline). This is because you're importing a service directly rather than an abstraction.

src/index.html
```diff
@@ -19,7 +19,7 @@
     <app-root>
         <div class="mt-5 d-flex justify-content-center">
             <div>
-                <img src="./images/logo-dark@2x.png" class="mb-4 logo" alt="Bitwarden">
+                <div id="loginLogo" class="mb-4 logo" alt="Bitwarden"></div>
```
eliykat, line 22: Is it right that this has changed from an `img` to a `div` even though it's still just to display an image?

src/app/settings/payment.component.ts
```diff
@@ -61,8 +49,11 @@ export class PaymentComponent implements OnInit {
     private stripeCardNumberElement: any = null;
     private stripeCardExpiryElement: any = null;
     private stripeCardCvcElement: any = null;
+    private theme: any = null;
+    private StripeElementStyle: any;
+    private StripeElementClasses: any;
 
-    constructor(private platformUtilsService: PlatformUtilsService, private apiService: ApiService) {
+    constructor(private platformUtilsService: PlatformUtilsService, private apiService: ApiService, private storageService: StorageService) {
```
eliykat, line 56: Probably need to bump this down to the next line. See the rule below:

Screen Shot 2021-08-05 at 3 44 20 pm

eliykat (changes requested):
You've done a great job here and I can see you've put a ton of work into this, so thank you!

I've left some more detailed comments below. This is purely code review, there might be further feedback on design elements, but that hopefully shouldn't affect the code structure and scaffolding too much. It looks like a lot of the work was just getting theming set up to begin with.

One random question I couldn't add to the file directly: should `logo-white@2x.png` be the same size as the existing `logo-dark@2x.png`? Currently white is 1500x300px but dark is 568x86px.

src/app/settings/payment.component.ts
dltmurphy: OKay I've added it above the webConstants one so both services are together unless it's better to keep the jslib separated by a new line from this import too?

src/index.html
dltmurphy: Yeah, I've changed it to be a background image which can be swapped out with the theme in the stylesheet

src/app/settings/payment.component.ts
eliykat: Yes, I think keep it separate from `webConstants` because it's coming from jslib. So the import groups here will look like:
* jslib abstractions
* jslib services (ConstantsService)
* web services (WebConstants)

src/scss/headers.scss
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+}
+
+.navbar-brand {
+    margin-bottom: -20px;
+    margin-top: -20px;
+}
+
+.page-header, .secondary-header {
```
eliykat, line 50: Most of this file relates to the navBar, which is not really a 'header' in the same way `.page-header` and `.secondary-header` is. I think we just need to redistribute this file among the other files.
* move this block (`.page-header...`) and the next block (`.secondary-header...`) into `base.scss`
* rename this file (with the styles that are left in it) `misc.scss`

src/scss/register-layout.scss
```diff
@@ -1,47 +1,57 @@
 .layout {
     &.enterprise2 {
-
```
eliykat, line 3: This can be moved to `pages.scss`, because it provides styling for a specific page (the register/new account page). Just put a comment before this block identifying the page it relates to, because it's not super obvious from the code itself.

src/scss/tables.scss
```diff
@@ -0,0 +1,124 @@
...
+}
+
+.table-hover tbody tr:hover {
+    @include themify($themes) {
+        background-color: themed('tableRowHover');
+        color: themed('tableColorHover');
+    }
+}
+
+.list-group-item {
```
eliykat, line 114: I don't think this relates to tables, please find somewhere else to put it (maybe just `misc` unless you can think of anywhere better?) (correct me if I'm wrong)

eliykat (changes requested):
The SCSS is looking much better after splitting it up. That said, I think a few categories are too granular and could be combined or changed:
* `app.scss` - rename to `pages.scss` (it provides page-specific styling, this is the wording used in the browser extension styles)
* add `typography.scss` to the end of `base.scss` (so you're just left with `base.scss`
* similarly, add `dropdowns.scss` to the end of `forms.scss`
* see additional feedback below

Unfortunately there's been some changes to `styles.scss` since your last commit, which indicates others have made changes to the styles in the meantime (my fault for not reviewing sooner). Please review the merge conflicts to see what's been changed, and then replicate those changes in the new location. Basically we need to make sure that we're not accidentally undoing (or blasting over) changes that others have made in the meantime.

dltmurphy:
I've moved them about, i just need to do some testing to make sure it all works and then will commit and push :)

src/app/services/services.module.ts
```diff
@@ -163,7 +163,11 @@ export function initFactory(): Function {
         htmlEl.classList.add('locale_' + i18nService.translationLocale);
         let theme = await storageService.get<string>(ConstantsService.themeKey);
         if (theme == null) {
-            theme = 'light';
+            theme = await platformUtilsService.getDefaultSystemTheme();
+            platformUtilsService.onDefaultSystemThemeChange(sysTheme => {
+                htmlEl.classList.remove('theme_light', 'theme_dark');
```
eliykat, line 168: There's a bit of a bug at the moment where this event handler is only registered when the app is started based on the theme at that point in time, so if you change the theme later it will not activate or deactivate the event handler.

To avoid this, I recommend always registering the event handler here, and then checking whether `theme == null` inside the callback.

src/images/loading.svg
```diff
@@ -1,6 +1,6 @@
 ﻿<svg xmlns="http://www.w3.org/2000/svg" width="100%" height="100%" viewBox="0 0 100% 100%">
-  <text fill="%23333333" x="50%" y="50%" font-family="\'Open Sans\', \'Helvetica Neue\', Helvetica, Arial, sans-serif"
+  <text fill="#333333" x="50%" y="50%" font-family="\'Open Sans\', \'Helvetica Neue\', Helvetica, Arial, sans-serif"
```
eliykat, line 2: I'm not sure what this does, is it required if you've created a separate `loading-white.svg`?

src/scss/variables.scss
```diff
@@ -0,0 +1,347 @@
...
+$input-placeholder-color: #B4B4B4;
+
+$table-accent-bg: rgba(#000000, .02);
+$table-hover-bg: rgba(#000000, .03);
+
+$modal-backdrop-opacity: 0.3;
+$btn-font-weight: 600;
+$lead-font-weight: normal;
+
+$grid-breakpoints: (
```
eliykat, line 68: This isn't referenced anywhere and also doesn't seem like it has valid values

src/scss/base.scss
```diff
@@ -0,0 +1,286 @@
+html {
+    font-size: 14px;
+}
+
+body {
+    min-width: 1010px;
+
+    &.layout_frontend {
+        @media (prefers-color-scheme: dark) {
```
eliykat, line 9: Do you need these 2 `@media` blocks? The `themify` mixin is intended to handle this. Also these will only read the system colour theme, not the selected colour theme.

src/scss/base.scss
```diff
@@ -0,0 +1,286 @@
...
+
+.overflow-hidden {
+    overflow: hidden;
+}
+
+.cursor-move {
+    cursor: move !important;
+}
+
+.bg-primary {
```
eliykat, line 145: Are all these uses of `!important` necessary? You should generally use specificity to resolve competing CSS rules, `!important` is considered bad practice and can make stylesheets difficult to maintain. Please try to refactor this document so that you're not using `!important` (unless it was already there before your changes)

src/scss/buttons.scss
```diff
@@ -0,0 +1,224 @@
...
+.badge-primary {
+    @include themify($themes) {
+        background-color: themed('badgePrimaryBackground');
+        color: themed('badgePrimaryText');
+    }
+
+    &:hover {
+        @include themify($themes) {
+            background-color: themed('badgePrimaryBackgroundHover');
+            color: themed('badgePrimaryText') !important;
```
eliykat, line 186: Again just double check that this is necessary

eliykat (changes requested):

.dccache
```diff
@@ -0,0 +1 @@
...mon\\src\\models\\domain\\organization.ts":[4375,1630500871300.1074,"ac222e63729d594d3e7d08c5feef3b3a88f633fac24a0fb441f07045d85484ac"],"c:\\Github\\Bitwarden\\web\\jslib\\common\\src\\models\\response\\profileResponse.ts":[2396,1630500871300.1074,"cb1633f953f8a238524c597324937abbb917c559f256c169c0554976c8618051"],"c:\\Github\\Bitwarden\\web\\jslib\\common\\src\\services\\i18n.service.ts":[5724,1630500871301.1033,"9b524f5c5aee2842dd8d278ce1812e6442bbfaed895318965683afce129a4e39"],"c:\\Github\\Bitwarden\\web\\jslib\\common\\src\\services\\policy.service.ts":[7284,1630500871301.1033,"23e5cb3d31de1cdf0399f23ae3d54be9d3400932b174f35e34f46176191c1ff0"],"c:\\Github\\Bitwarden\\web\\jslib\\electron\\src\\services\\electronPlatformUtils.service.ts":[5925,1630500871303.1099,"4d11a1fc666c3545b2a65edc9f7c55740ffe810279bec6cbc80bae74d5596a92"],"c:\\Github\\Bitwarden\\web\\jslib\\node\\src\\cli\\commands\\login.command.ts":[17181,1630500871304.104,"dc3170d459c3af70894b24ba2781bfd2979013f921e7001d1167622c6f7ca6f2"],"c:\\Github\\Bitwarden\\web\\jslib\\node\\src\\cli\\services\\cliPlatformUtils.service.ts":[3603,1630500871304.104,"9073bbf38fef3e22be596058b1781c24fd29bef11bc7ebf6808cec430f5a637c"]}
```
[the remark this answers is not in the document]
dltmurphy: My bad 😂 Will get that removed on the next commit

src/scss/variables.scss
dltmurphy: This is just what was in the styles.scss originally. I haven't removed any of the previously set variables, I can remove them if you want?

src/scss/base.scss
dltmurphy: This part is the initial loading screen. In the previous PR it was requested by somebody to add it in as if you're in dark theme you'd get a flash of white when loading the vault

src/images/loading.svg
dltmurphy: This just replaces the %23 code with the # as we aren't using the SVGs with fragment identifiers that i can see. I can edit both of them so they use %23 if preferred?

src/app/settings/options.component.ts
```diff
@@ -80,6 +89,16 @@ export class OptionsComponent implements OnInit {
         await this.stateService.save('enableGravatars', this.enableGravatars);
         await this.storageService.save('enableFullWidth', this.enableFullWidth);
         this.messagingService.send('setFullWidth');
+        if (this.theme !== this.currentTheme) {
+            let effectiveTheme = this.currentTheme = this.theme;
+            await this.storageService.save('theme', this.theme);
+            const htmlEl = window.document.documentElement;
+            htmlEl.classList.remove('theme_dark', 'theme_light');
```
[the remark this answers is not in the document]
dltmurphy: Just moved `htmlEl.classList.remove('theme_dark', 'theme_light');` to above the `add('theme_'` line

src/scss/variables.scss
eliykat: Oh ok, it's a little tricky to tell what's new and what's old. Let's just leave it alone then, we're changing enough and I'm going to presume that it was there for a reason.

src/scss/base.scss
eliykat: Resolved, see `index.html` and `misc.scss`

eliykat:
I'm requesting a final review from the team because I have contributed code to this as well and someone with fresh eyes should approve it. Comments:

* Depends on some refactors in this jslib PR: https://github.com/bitwarden/jslib/pull/497
* `index.html` contains inline javascript to set the theme before the Angular app loads. Obviously inline js is ugly, but we'd need to add `babel-loader` to our build chain if we wanted it in a separate file, which didn't seem worth it just for this.
* the registration page has a special marketing campaign layout if you pass `?layout=enterprise2` in the URL params. This is not properly themed for dark theme, however my assumption is that if you're clicking on an email campaign link to this page, you don't have a Bitwarden account and it'll default to the light theme. And conversely, if you have an account and opted in to the dark theme, you won't be visiting this page. If we want to be totally sure, I could force light theme for this page, but I haven't done that for now.
* the `base.scss` has a lot of `!important` flags in it, which is bad practice, however we need to override Bootstrap theming which also uses `!important`, so I couldn't find a more elegant way to do it.

Other notable changes:
* split scss into several files and build out themes using the standard `themify` mixin (previously web did not have any support for this at all)
* add selector to `options.component` to allow the user to select themes
* add a `themed-image.component` for a couple of images that need to switch dynamically between light and dark mode versions
* refactor `platformUtilsService` to match the jslib refactor linked above

Hinton:
@eliykat Won't we still need to theme or at least disable the dark mode for `?layout=enterprise2` since they could have the `prefer-dark-mode`.

src/services/webPlatformUtils.service.ts
```diff
@@ -283,11 +288,24 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
...
     }
 
-    getDefaultSystemTheme() {
-        return Promise.resolve(null as 'light' | 'dark');
+    getDefaultSystemTheme(): Promise<ThemeType.Light | ThemeType.Dark> {
+        return Promise.resolve(this.prefersColorSchemeDark.matches ? ThemeType.Dark : ThemeType.Light);
+    }
+
+    async getEffectiveTheme(): Promise<ThemeType.Light | ThemeType.Dark> {
+        const theme = await this.storageService().get<ThemeType>(ConstantsService.themeKey);
```
Hinton, line 296: Do we need to default to `'system'` here?

Hinton (changes requested):
I took a fairly quick glance at the `scss` code but focused mainly on the other areas. Overall great work! A few minor comments regarding if we should default to `system`, some suggestions on how we can either avoid the `ThemedImageComponent` or replace it with a pipe.

src/index.html
```diff
@@ -17,9 +17,21 @@
 
 <body class="layout_frontend">
     <app-root>
+        <script>
+            (function () {
+                // Set theme before app loads - prevents flash of white when using dark mode
+                let theme = window.localStorage.getItem('theme')?.replace(/"/g, '');
```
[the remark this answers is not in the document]
eliykat: This was a design decision discussed with Danielle. Concerns were:
* surprising existing users who wake up one day to find their web vault in dark mode
* new users being presented with dark colours on their first interaction with Bitwarden, rather than our official brand colours
* the marketing campaign registration page not being designed with dark theme in mind - this will require some coordination with marketing, and since there will be some registration improvement work on the 2022 roadmap, we'll revisit it at that time

I know this makes it slightly out of step with other clients (which default to `system`), but I don't think that's a big deal.

.gitignore
```diff
@@ -13,3 +13,4 @@ dist/
 build/
 !dev-server.shared.pem
 config/local.json
+.dccache
```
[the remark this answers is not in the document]
eliykat: Reverted

src/services/webPlatformUtils.service.ts
eliykat: See response above

src/app/services/services.module.ts
```diff
@@ -161,11 +163,17 @@ export function initFactory(): Function {
...
         const htmlEl = window.document.documentElement;
         htmlEl.classList.add('locale_' + i18nService.translationLocale);
-        let theme = await storageService.get<string>(ConstantsService.themeKey);
-        if (theme == null) {
-            theme = 'light';
-        }
+
+        const theme = await platformUtilsService.getEffectiveTheme();
         htmlEl.classList.add('theme_' + theme);
+        platformUtilsService.onDefaultSystemThemeChange(async sysTheme => {
```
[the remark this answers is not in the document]
eliykat: `index.html` only sets the initial theme, it doesn't set up the `onDefaultSystemThemeChange` handler.

Your point still applies to setting the initial theme, we could delete it from lines 167-168 in this file because it's handled in `index.html`. However, `index.html` is more brittle because it can't hook into the theme enums or the `platformUtilsService` methods, it's a bit of an anomaly. I'm not sure whether it's worth keeping this here as a backup, even though the duplication is ugly. What do you think?

src/index.html
```diff
@@ -17,9 +17,21 @@
 
 <body class="layout_frontend">
     <app-root>
+        <script>
```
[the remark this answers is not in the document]
eliykat: I put it in the `app-root` tag because then it'll disappear from the DOM once the app loads in. It's really part of the loading elements contained in the `app-root` tag, not something that should hang around for the lifetime of the app. (Unless we decided to remove the theme setting logic from `services.module.ts` per the discussion below, then this should move into the `<header>` tags.)

I know that `<script>` should generally be in `<header>` though, let me know if you still think that's the way to go and I'll change it.

src/app/accounts/login.component.html
```diff
@@ -1,7 +1,9 @@
 <form #form (ngSubmit)="submit()" [appApiAction]="formPromise" class="container" ngNativeValidate>
     <div class="row justify-content-md-center mt-5">
         <div class="col-5">
-            <img src="../../images/logo-dark@2x.png" class="logo mb-2" alt="Bitwarden">
```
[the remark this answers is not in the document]
eliykat: I thought that we used `img` here because it's semantic, it's not just a background image, it's the title/header image for the page and it should get picked up by screen readers and other assistive technologies.

We use CSS `background-image` for `index.html` but I consider that a hack that I didn't really want to replicate elsewhere (even though that would at least be consistently hacky 😆 )

I like your idea of a pipe though, I'll investigate that.

src/app/components/themed-image.component.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { ConstantsService } from 'jslib-common/services/constants.service';
+
+import { ThemeType } from 'jslib-common/enums/themeType';
+
+@Component({
+    selector: 'app-themed-image',
+    templateUrl: 'themed-image.component.html',
+})
+export class ThemedImageComponent implements OnInit, OnDestroy {
```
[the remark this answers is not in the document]
eliykat: I agree this could be a better solution, I probably under-utilize pipes in Angular. I'll give it a go.

src/app/accounts/login.component.html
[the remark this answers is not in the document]
eliykat: Update: I realised that this is a solved problem in the desktop client, so I used the same solution here. We can use scss to set the image, using the `content` attribute. That keeps it really simple while also being able to keep the semantic `img` tag with its `alt` atribute for accessibility.

src/app/components/themed-image.component.ts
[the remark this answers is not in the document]
eliykat: As explained in my other comment, this component has now been removed.

src/index.html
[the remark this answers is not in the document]
Hinton: Fair enough, do we have a card in Asana to track this?

src/scss/variables.scss
```diff
@@ -0,0 +1,349 @@
...
+        foregroundColor: $darkDarkBlue1,
+        headerColor: $darkBlue1,
+        iconColor: #777777,
+        iconHover: $darkGrey2,
+        imgFilter: invert(1) grayscale(1),
+        inputBackgroundColor: $darkDarkBlue1,
+        inputBorderColor: $darkGrey1,
+        inputDisabledBackground: $darkBlue2,
+        inputDisabledColor: $darkGrey1,
+        inputPlaceholderColor: $darkGrey1,
```
Hinton, line 301: We should use the variables here as well, similar to light theme.

dltmurphy: Ah my bad. Updated these now

src/index.html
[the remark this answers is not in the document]
eliykat: I'll follow up with Danielle and make sure it's tracked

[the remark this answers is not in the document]
eliykat: https://app.asana.com/0/1153292148278596/1201082114808085/f

Hinton (dismissed):
Looks good to me! Good work @dltmurphy and @eliykat.

Hinton (approved):

eliykat (approved):

PR #1022, opened 2021-06-09
Fix #1020 - XSS via innerHTML property

### LINKED ISSUE #1020: Possible DOMXSS in webauthn-fallback.ts
Hi!

The Snyk Code scanner I used to review some of the code of this project mentioned these two code sections, where input is being directly used to influence the `innerHTML`
https://github.com/bitwarden/web/blob/82ec4b12f7393c4479df494d7e82e38da0d275bc/src/connectors/webauthn-fallback.ts#L117

https://github.com/bitwarden/web/blob/82ec4b12f7393c4479df494d7e82e38da0d275bc/src/connectors/webauthn-fallback.ts#L107

In case the `message` parameter can be influenced by an attacker, this could lead to an XSS issue. Looking at this, it does not seem too dangerous, but at the same time it would be better to encode the string before putting it into the `innerHTML` or put the content into `textContent` instead. 
More details: https://stackoverflow.com/questions/30661497/xss-prevention-and-innerhtml

cscharf:
Closes #1020 

## Overview

While an extremely low risk, it's best practice to not use `.innerHTML` for assignment of values dynamically or coming from unknown origins, in this case within a method where the caller may not be known (OOP concept of encapsulation). It would also appear that in this instance this method is only called in exactly 1 place so if we need special handling of HTML in the future here (e.g. this change isn't entirely _correct_) then we can split the message up and handle it accordingly.

* **webauthn-fallback.ts ** - Assign values to `. textContent` instead of `.innerHTML` to remove the possibility of injecting `<script/>` or other XSS techniques.
* **en/messages.json** - Removed the HTML from the translation string. It may not look as pretty on-screen, but for now changes the only translation string that contained markup to no longer contain markup.

### DIFF
--- a/src/connectors/webauthn-fallback.ts
+++ b/src/connectors/webauthn-fallback.ts
@@ -104,7 +104,7 @@ async function initWebAuthn(obj: any) {
 function error(message: string) {
     const el = document.getElementById('msg');
     resetMsgBox(el);
-    el.innerHTML = message;
+    el.textContent = message;
     el.classList.add('alert');
     el.classList.add('alert-danger');
 }
@@ -114,7 +114,7 @@ function success(message: string) {
 
     const el = document.getElementById('msg');
     resetMsgBox(el);
-    el.innerHTML = message;
+    el.textContent = message;
     el.classList.add('alert');
     el.classList.add('alert-success');
 }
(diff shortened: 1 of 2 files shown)


Hinton (approved):
Slightly sad we won't have the well formatted message but yea better to use `textContent`.

bt-nia:
Thanks everyone! Glad to see this being added. I understand this is really low risk, but appreciate sticking to best practices to ensure this can not be exploited in what ever weird exploit chain. 

Cheers! :)

PR #1024, opened 2021-06-10
Add bump-jslib Github action

eliykat:
## Objective

As discussed in Slack, add a new Github action that lets us bump jslib without having to dive into the command line.

Behaviour:
* must be manually triggered from the "Actions" tab
* if run against master, it will create a new branch, bump jslib on that branch, and submit a PR against master (or update the existing PR if there's already one pending)
* if run against a feature branch, it will bump jslib on that branch and take no further action

## To do

* Permissions should be set so that only Bitwarden org members can run it - @joseph-flinn how does this work?
* Currently it's got me marked as the reviewer so that we don't spam the team with review requests while testing. This will need to be changed back to `bitwarden/dept-engineering` once we're happy with it.
* Once it's finalised, add to all other repos as well.

@joseph-flinn Thanks for your help with this, please feel free to push to this branch if you want to make any changes yourself, or leave any feedback and I'll make them.

### DIFF
--- a/.github/workflows/bump-jslib.yml
+++ b/.github/workflows/bump-jslib.yml
@@ -80,11 +80,10 @@ jobs:
             gh pr create --title "Auto bump jslib" \
               --body "$BODY_TEXT" \
               --reviewer eliykat
-            
             echo "[*] Submitted PR against master branch"
           else
-            gh pr comment \
-              --body "$BODY_TEXT"
+            gh pr comment --body "$BODY_TEXT"
+            echo "[*] Updated existing PR"
           fi
         env:
           GIT_LOG: ${{ steps.bump.outputs.git-log }}
--- a/.github/workflows/bump-jslib.yml
+++ b/.github/workflows/bump-jslib.yml
@@ -1,0 +1,74 @@
+name: Bump jslib
+
+on:
+  workflow_dispatch:
+
+jobs:
+  bump-jslib:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Setup git config
+        run: |
+          git config user.name = "GitHub Action Bot"
+          git config user.email = "<>"
+
+      - name: Bump jslib and commit
+        id: bump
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-jslib.yml
```diff
@@ -0,0 +1,74 @@
...
+on:
+  workflow_dispatch:
+
+jobs:
+  bump-jslib:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
```
joseph-flinn, line 12:
```suggestion
        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
```

To mitigate some risks of some supply chain attacks, we should use the commit hash of the Action instead of the version

.github/workflows/bump-jslib.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Bump jslib and commit
+        id: bump
+        run: |
+          TARGET_BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/})
+          NEW_BRANCH_NAME="auto-bump-jslib"
+
+          if [ "$TARGET_BRANCH_NAME" = "master" ]; then
+            git checkout -b $NEW_BRANCH_NAME
+            echo "[*] Checked out new branch $NEW_BRANCH_NAME"
+          fi
```
joseph-flinn, line 28: Any issues with running this when `auto-bump-jslib` already exists in an open PR from a previous run?

joseph-flinn (changes requested):
By default, GitHub Actions can only be manually ran by members that have Contributor permissions or higher on the repo.

.github/workflows/bump-jslib.yml
eliykat: Yep - have now updated and refactored to allow for this. If there is an existing PR, it'll push to that branch instead and add a comment with the `git log` output.

joseph-flinn (approved):
It looks good! Great job!

PR #1025, opened 2021-06-10
Ensure we only select all visible users

Hinton:
## Objective
When clicking select all in organization people list, it would incorrectly select from all users instead of only the visible users.

### DIFF
--- a/src/app/organizations/manage/people.component.ts
+++ b/src/app/organizations/manage/people.component.ts
@@ -37,6 +37,8 @@ import { OrganizationUserStatusType } from 'jslib-common/enums/organizationUserS
 import { OrganizationUserType } from 'jslib-common/enums/organizationUserType';
 import { PolicyType } from 'jslib-common/enums/policyType';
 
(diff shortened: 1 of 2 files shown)


src/app/organizations/manage/people.component.ts
```diff
@@ -545,11 +548,14 @@ export class PeopleComponent implements OnInit {
         if (select) {
             this.selectAll(false);
         }
-        const selectCount = select && this.users.length > MaxCheckedCount
+
+        const filteredUsers = this.searchPipe.transform(this.users, this.searchText, 'name', 'email', 'id');
```
cscharf, line 552: How is the list itself actually getting filtered (before select all is clicked)? Isn't there a simpler or more direct way to get that list of items vs. re-filtering/re-searching the full list?

Hinton: The list is filtered in the template using
`<ng-container *ngIf="!loading && (isPaging() ? pagedUsers : users | search:searchText:'name':'email':'id') as searchedUsers">`.

cscharf (approved):

PR #1026, opened 2021-06-10
Refactor accept-organization and accept-emergency

Hinton:
## Objective
Since Provider (#1014) adds another `accept` flow, and will require yet another before it's feature complete, I decided to take this moment to refactor the accept components to minimize the required code. While at it, I also replaced the previous states with the more generic `loginRedirect`.

### Testing Considerations
We should do a regression sweep on accept emergency and organization flows.

### DIFF
--- a/src/app/accounts/accept-emergency.component.ts
+++ b/src/app/accounts/accept-emergency.component.ts
@@ -24,7 +24,7 @@ export class AcceptEmergencyComponent extends BaseAcceptComponent {
 
     name: string;
 
-    protected requiredParameters: string[] = ['name', 'email', 'token'];
+    protected requiredParameters: string[] = ['id', 'name', 'email', 'token'];
     protected failedShortMessage = 'emergencyInviteAcceptFailedShort';
     protected failedMessage = 'emergencyInviteAcceptFailed';
 
--- a/src/app/accounts/accept-emergency.component.ts
+++ b/src/app/accounts/accept-emergency.component.ts
@@ -1,7 +1,4 @@
-import {
-    Component,
-    OnInit,
-} from '@angular/core';
+import { Component } from '@angular/core';
 import {
     ActivatedRoute,
     Router,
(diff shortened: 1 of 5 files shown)


cscharf (dismissed):

src/app/accounts/accept-emergency.component.ts
```diff
@@ -17,77 +14,47 @@ import { I18nService } from 'jslib-common/abstractions/i18n.service';
...
+
     name: string;
-    email: string;
-    actionPromise: Promise<any>;
 
-    constructor(private router: Router, private toasterService: ToasterService,
-        private i18nService: I18nService, private route: ActivatedRoute,
-        private apiService: ApiService, private userService: UserService,
-        private stateService: StateService) { }
+    protected requiredParameters: string[] = ['name', 'email', 'token'];
```
Hinton, line 27: Missing id.

cscharf (approved):

BlackDex:
Hello there,

I'm sorry if this comment needs to be somewhere else, but i think the issue i experience is related to this PR.

I am one of the contributors of the Vaultwarden project. And when i tried to use a web build which had this PR included it doesn't seem to work without issues. Now i don't know if using master causes some strange issues which you already know.

What happens is the following. When i create an invite for a new user and that users is using that link all goes fine up-until the part where it doesn't seem to trigger the master-password policy check. Also, the token isn't sent with the request anymore when the user presses the submit button.

If i use the specific commit hashes from the `v2.21.0` or `rc` this issue doesn't seem to be there, since it doesn't have this specific commit in there chain.

Hinton:
@BlackDex Hmm, I can't seem to replicate the issue. I've tried joining an organization with and without master password reset enabled, and setting up emergency access. Using the latest master of both web vault and server. I even tried using `npm run build:prod:watch` which runs against the current production server.

Could you check the browser console and/or network tab if there is any specific error which occurs and which request that fails?

BlackDex:
@Hinton, thanks for the reply.
It isn't the master password reset in this case. I haven't even tested that out yet.
It is regarding the password policy, so how much chars, the weakness allowed etc... I just tested the GH Action build images also and extracted the the `/app` folder from `v.2.21.0`, `rc` and `dev` docker hub images of `bitwarden/web`. And only `dev` fails to get the password policy.

It is regarding a call to this endpoint: `https://my.domain.tld/api/organizations/ORG-ID/policies/token?token=very-long-token`
That one isn't triggered for me on the `dev` extracted from the docker hub image, or if i build the web-vault my self based upon master.

If i build the web-vault based upon the `v2.21.0` (`c852d536ea4fe3e2a81880a77cd688f77583b46a`) or the `rc` (`f3c1c68d0e372490d2ce3afacfaa7e47f4b977f4`) hashes they do work, since those do not yet contain this specific patch.

I even build a version based upon the commit hash right before this PR got merged (`b8aff13ce2a186af81efd81ae00702df225b233c`) also build the a version of the hash of this PR merged (`14e9784297fd07b251322277cdd0bf7411252f02`). The `b8aff13` works, the `14e9784` doesn't.

It looks like the function `getPoliciesByToken` isn't called anymore: https://github.com/bitwarden/web/blob/e7f3d977c4e1c2ed2c978e89c8adc2ad049d026e/src/app/accounts/register.component.ts#L96

There are no errors in the dev-console.

So the steps i do to test this is the following:
1. Create an invite via the organization web-vault which generates an e-mail with a link.
2. Open an Private/Incognito browser window to test with and open the developer console before i paste the URL.
3. Open the URL, click on the `Create Account` button.
4. Check the `Network` tab to see if the endpoint is called.

With the `14e9784` commit i see this, which doesn't show the password policies:
image

With the `b8aff13` commit i see this, which does show me the policies:
image

If  you need more information please let me know.

Hinton:
@BlackDex Thanks for the additional information, seems I was testing the wrong things. Yea that looks like a bug introduced by this PR.

PR #1027, opened 2021-06-11
[Reset Password] Custom Permission pairing

vincentsalucci:
## Code Changes
- **user-add-edit.component.html**: Added method call that fires during the `(change)` action
- **user-add-edit.component.ts**: Created new `handleDependentPermissions` method that will auto-select the `Manage Users` permission if necessary
- **messages.json**: Added informational string during auto-selection

## Screenshot
Screen Shot 2021-06-11 at 9 17 39 AM

### DIFF
--- a/src/app/organizations/manage/user-add-edit.component.html
+++ b/src/app/organizations/manage/user-add-edit.component.html
@@ -172,7 +172,8 @@
                                 <div class="form-group mb-0">
                                     <div class="form-check mt-1 form-check-block">
                                         <input class="form-check-input" type="checkbox" name="manageUsers"
-                                            id="manageUsers" [(ngModel)]="permissions.manageUsers">
+                                            id="manageUsers" [(ngModel)]="permissions.manageUsers"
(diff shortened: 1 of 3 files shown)


kspearrin:
Where did this new blue toat message come from? I have never seen us using blue toast in the past. I think on other clients an info toast is a black/gray look.

vincentsalucci:
> Where did this new blue toat message come from? I have never seen us using blue toast in the past. I think on other clients an info toast is a black/gray look.

Its just the standard use of `platformUtilsService.showToast('info'...)`. We use it in a handful of places in the `web` client. For example, when you copy a generated password using the Password Generator. You may be thinking of the `app-callout` which is black/gray for the `info` type, I believe.

kspearrin:
Ok, I think we just need a task to style that in line with other clients, for example, the browser extension. The info toast there is gray.

cscharf (approved):
This looks good Vince, we can tackle the info styling in other PR.

PR #1034, opened 2021-06-16
Add show/hide button to password prompt

Hinton:
## Objective
Add a show/hide button to the password field.

### Testing Considerations
We should do a regression sweep of the dialog.

### Screenshots
chrome_STpNOaB08E

### DIFF
--- a/src/services/webPlatformUtils.service.ts
+++ b/src/services/webPlatformUtils.service.ts
@@ -224,11 +224,28 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
 
     async showPasswordDialog(title: string, body: string, passwordValidation: (value: string) => Promise<boolean>):
         Promise<boolean> {
+        
+        const html = `
+        ${body}
+        <div class="form-group mt-2">
+            <label for="masterPassword">${this.i18nService.t('masterPass')}</label>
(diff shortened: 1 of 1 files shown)


src/services/webPlatformUtils.service.ts
```diff
@@ -224,24 +224,61 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
 
     async showPasswordDialog(title: string, body: string, passwordValidation: (value: string) => Promise<boolean>):
         Promise<boolean> {
+
+        const html = `
```
cscharf, line 228: It really feels like we need look at switching to using templates or components to pop into these dialogs vs. embedding HTML in strings and logic around where the dialog is shown (vs. encapsulating the dialog into its own component w/ logic and display).

cscharf (approved):

src/services/webPlatformUtils.service.ts
Hinton: Yea I felt that way as well. I will spend a bit of time looking into using the `repromptService` to show a dialog instead of using SweetAlert.

Hinton: Decided to merge this for now, and postpone move to modals.

PR #1035, opened 2021-06-16
Must await to get a value

MGibson1:
# Overview

Fix issue where we aren't awaiting `get` to return its value rather than the promise of one.

FYI, this is a regression item and will be cherry-picked into `rc` upon merger to master. Please evaluate accordingly.

# Testing.

This regression manifests as a failure to lock and failure to decrypt the web vault upon reloading.


cscharf (approved):
This `has` looks good.

PR #1038, opened 2021-06-17
Force download of file regardless of type

OJ:
The file download feature in Bitwarden Send currently treats PDF files as a special case by not forcing the file to be downloaded. By the looks of it the reasoning is to make use of the built-in PDF viewer that comes with the browser.

Based on my use of the application, and our extensive sharing of PDFs with non-technical people, this special case is a genuine usability problem. Almost every person we have shared a PDF document with has thought that the download feature hasn't worked because of the built-in popup blocker.

The code in question makes an exemption for Safari, so it's not an issue in that case, but for every other browser this is proving to be a concern.

From a usability point of view, I do not think that PDF files should be treated any differently. When a user clicks a link that says "Download File", they would expect a file to be downloaded. That's certainly what I would expect. I also personally dislike it when applications try to be too smart in cases like this and try to use built-in browser features for certain file types.

Attempting to download a file in Bitwarden Send should result in the file being downloaded. If the user has their browser configured to view PDF files, then that's what will happen when they try to open it. If they do not, then their other chosen viewer will be invoked. This option is still under the control of the user even when the file has been downloaded first.

This small change removes the code that treats PDF files as a special case. Instead the files are downloaded directly just like any other file type. This consistency is a win for useability and avoids the problem of popup blocker getting in the way of downloading PDF files.

I checked to see where else in the code this download feature is used and found the following:

```
./src/app/settings/user-subscription.component.ts:        this.platformUtilsService.saveFile(window, licenseString, null, 'bitwarden_premium_license.json');
./src/app/organizations/manage/events.component.ts:        this.platformUtilsService.saveFile(window, data, { type: 'text/plain' }, fileName);
./src/app/organizations/settings/download-license.component.ts:            this.platformUtilsService.saveFile(window, licenseString, null, 'bitwarden_organization_license.json');
```

The above cases are all non-PDF files, so the change does not impact them at all.

I'm sure there were good reasons for putting this feature in place to begin with. I just feel that treating any file as a special case is a bad idea, removes control from the user and results in unexpected behaviour. It's especially bad when downloads just don't work because popups are silently blocked.

I welcome any discussion on the topic. Thanks very much for your time.

### DIFF
--- a/src/services/webPlatformUtils.service.ts
+++ b/src/services/webPlatformUtils.service.ts
@@ -103,10 +103,8 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
         let blob: Blob = null;
         let type: string = null;
         const fileNameLower = fileName.toLowerCase();
-        let doDownload = true;
         if (fileNameLower.endsWith('.pdf')) {
             type = 'application/pdf';
-            doDownload = false;
         } else if (fileNameLower.endsWith('.xlsx')) {
             type = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet';
         } else if (fileNameLower.endsWith('.docx')) {
@@ -137,11 +135,7 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
             navigator.msSaveBlob(blob, fileName);
         } else {
             const a = win.document.createElement('a');
-            if (doDownload) {
-                a.download = fileName;
-            } else if (!this.isSafari()) {
-                a.target = '_blank';
-            }
+            a.download = fileName;
             a.href = URL.createObjectURL(blob);
             a.style.position = 'fixed';
             win.document.body.appendChild(a);


cscharf:
See: https://github.com/bitwarden/web/pull/888 for background/reference regarding the Safari issue. As far as functionality goes, I tend to agree that a download should be kept a download. I don't remember if there was explicit browser behavior issues with setting one way or the other given Safari and the blob data technique used to prompt the file download. Assigning to @MGibson1 for when he has time to do a quick review.

Asana Ref: https://app.asana.com/0/1169444489336079/1200050082951941/f

MGibson1 (changes requested):
@OJ, Thanks for taking this on. You aren't _quite_ correct that this has no other implications beyond Send. This same method is used to download attachments from Cipher items. I don't think we want to change the download/open behavior there, but we agree that the behavior _should_ change for send.

So, I think the way to handle this is to change the method signature of `saveFile` to include an option to `forceDownload`, which will default false, but we'll set to true for the Send `access` page.

OJ:
I didn't see any other refs in the code outside of send other than the ones shown above and they all seem to force a content type that isn't PDF. Would you mind pointing me at where I missed it?

Thank you!

MGibson1:
Sure, it's here: https://github.com/bitwarden/jslib/blob/master/angular/src/components/attachments.component.ts#L149, in the [jslib](bitwarden/jslib) repository.

PR #1042, opened 2021-06-21
add captcha connector

kspearrin:
Adds support for captcha connector that can be consumed via iframe from other Bitwarden clients for solving HCaptcha challenges.

### DIFF
--- a/src/connectors/captcha.html
+++ b/src/connectors/captcha.html
@@ -7,7 +7,7 @@
     <script src="https://hcaptcha.com/1/api.js"></script>
 </head>
 
-<body style="background: transparent;">
+<body>
     <div id="captcha"></div>
 </body>
 
--- a/src/connectors/captcha.html
+++ b/src/connectors/captcha.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset="utf-8" />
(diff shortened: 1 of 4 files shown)


src/connectors/captcha.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset="utf-8" />
+    <title>Bitwarden Captcha Connector</title>
+    <script src="https://hcaptcha.com/1/api.js"></script>
+</head>
+
+<body style="background: transparent;">
```
addisonbeck, line 10:
```suggestion
<body>
```

src/connectors/captcha.scss
```diff
@@ -0,0 +1,5 @@
+body {
+    min-width: 0px !important;
+    padding: 0;
+    margin: 0;
+}
```
addisonbeck, line 5:
```suggestion
body {
    min-width: 0px !important;
    padding: 0;
    margin: 0;
    background: transparent;
}
```

addisonbeck (commented):
Minor nit, just seems like you don't need to do an inline style here.

kspearrin:
> 
> 
> Minor nit, just seems like you don't need to do an inline style here.

Accepted your suggestions.

addisonbeck (approved):

PR #1045, opened 2021-06-24
Refactor OrganizationPlansComponent

Hinton:
## Objective
The `OrganizationPlansComponent` is difficult to follow and has three different flows depending on the criteria. I've refactored `submit()` it into a few smaller functions which better describe what happens, and cleaned up the conditional flow to avoid multiple if statements.

This is a preparatory step towards being able to add yet another flow for providers.


addisonbeck (approved):
Nice!

PR #1048, opened 2021-06-25
maxAccessCountDesc locale string update

choutkamartin:
Hey, I noticed that the string doesn't have a capitalized "Send" word.

It a bit messes up with the Crowdin comparing system, as both browser and desktop versions have "Send" capitalized correctly - while mobile and web don't.

I will make a PR on the mobile version too.


eliykat:
Yes, Send should definitely be capitalized. Thank you!

eliykat (approved):

PR #1055, opened 2021-07-01
[Reset Password] BUG Org Keys backfill force sync

vincentsalucci:
## Code Changes
- **people.component.ts**: Force a one time `sync` operation that will replace all of the Organizational data in local storage with the latest from the server.

## Notes
- This is a bit quick & dirty but is how we currently handle changing organizational data/refreshing the local storage. A potentially cleaner way would be to change the data locally via a successful API response and save those changes to local storage. Unfortunately the only way to update /any/ organizational data is by using API responses directly from the server en mass (no singular update organization method, ex: ProfileResponse from Sync).


cscharf (approved):

cscharf:
Please cherry-pick after merge, FYI @kendratodd

PR #1059, opened 2021-07-02
chore: radio instead of dropdown (refactor)

knhash:
**Feature name:**

```
Replace the drop down for password/passphrase option with a radio button choice
```

**Feature Description**

There are only two options in the drop down, would it be better to replace it with a radio button for choice?

- Aids in discoverability: I completely missed the option for a whole year before I thought to click on the option and see what it contains.
- Less number of interactions: It would take 1 less click to get the same result

Community Forum post: https://community.bitwarden.com/t/radio-choice-password-passphrase/30636

### DIFF
--- a/src/app/tools/password-generator.component.html
+++ b/src/app/tools/password-generator.component.html
@@ -10,15 +10,12 @@
     </div>
 </div>
 <div class="form-group">
-    <div class="form-check form-check-inline">
-        <input id="generate-password" name="type" value="password" class="form-check-input" type="radio"
-            (change)="saveOptions()" [(ngModel)]="options.type">
-        <label for="generate-password" class="form-check-label">{{'password' | i18n}}</label>
-    </div>
-    <div class="form-check form-check-inline">
-        <input id="generate-passphrase" name="type" value="passphrase" class="form-check-input" type="radio"
-            (change)="saveOptions()" [(ngModel)]="options.type">
-        <label for="generate-passphrase" class="form-check-label">{{'passphrase' | i18n}}</label>
+    <div class="form-check form-check-inline" appBoxRow name="PassTypeOptions" *ngFor="let o of passTypeOptions">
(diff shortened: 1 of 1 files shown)


src/app/tools/password-generator.component.html
```diff
@@ -10,15 +10,12 @@ <h1>{{'passwordGenerator' | i18n}}</h1>
...
-    <div class="form-check form-check-inline">
-        <input id="generate-password" name="type" value="password" class="form-check-input" type="radio"
-            (change)="saveOptions()" [(ngModel)]="options.type">
-        <label for="generate-password" class="form-check-label">{{'password' | i18n}}</label>
-    </div>
-    <div class="form-check form-check-inline">
-        <input id="generate-passphrase" name="type" value="passphrase" class="form-check-input" type="radio"
-            (change)="saveOptions()" [(ngModel)]="options.type">
-        <label for="generate-passphrase" class="form-check-label">{{'passphrase' | i18n}}</label>
+    <div class="form-check form-check-inline" appBoxRow name="PassTypeOptions" *ngFor="let o of passTypeOptions">
```
eliykat, line 13: * The `appBoxRow` directive is not used in web, please remove
* The `name` will be identical for both `<div>` elements generated by `*ngFor`. I don't think you need the `name` for anything, please remove

eliykat (changes requested):
A couple of minor changes, then we can merge.

Sorry to be confusing, jslib has now been updated on `web/master` and includes your changes. Could you please revert the `jslib` update in this PR so that it only contains your changes to `password-generator.component.html`. If that causes you any problems, just let me know and I'll sort it out.

Thanks!

src/app/tools/password-generator.component.html
knhash: @eliykat Removed the stuff, reverted jslib commit. Tested it out, works good.

eliykat (approved):

PR #1066, opened 2021-07-05
Bump braintree to 1.30.1

Hinton:
## Objective
Solves an issue where the braintree dropin gets stuck in an infinite spinner if `paymentOptionPriority` is set. https://github.com/braintree/braintree-web-drop-in/blob/main/CHANGELOG.md

Note: This PR will be cherry picked to `rc` branch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "bitwarden-web",
-      "version": "2.20.4",
+      "version": "2.21.1",
       "hasInstallScript": true,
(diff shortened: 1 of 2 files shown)


vincentsalucci (approved):
@Hinton Can you please create a ticket in Asana for this (if one doesn't already exist) to keep track of testing for QA. Especially because we want this in the 'rc' branch. Might require a bit of regression with payment methods other than stripe.

Hinton:
@vincentsalucci Created a ticket in Asana, Braintree is only used for PayPal.

Issue #838, opened 2021-02-19
Emoji Organization Names break icon generation


roleohibachi:
## Describe the Bug

This is on a self-hosted server instance. The server is using a unicode-compliant locale.

An icon to represent an Organization is usually generated from the first two letters of the name. This is done in-browser as a "data:image/svg+xml;base64" rather than creating a file on the server's disk. However, if the organization name contains an emoji, the function to generate this "data:image" blob instead returns "undefined", which is understood by the browser as a file hosted at "example.com/undefined". This file doesn't exist, however the function keeps trying to generate and load it. This causes a "flickering" behavior of multiple page elements in Firefox and Chrome (and likely others). 

## Steps To Reproduce

1. Name an organization "🐛 my org"
2. save and refresh the page

## Expected Result

No page elements should "flicker" or move unexpectedly, even when errors occur. 

Furthermore, an icon with an emoji in it would be make sense.

## Actual Result

"Flickering" behavior of multiple page elements in Firefox and Chrome (and likely others). 

## Screenshots or Videos

https://user-images.githubusercontent.com/18207889/108556633-4fed2f00-72c5-11eb-9e08-511033b0d600.mp4

## Environment

- Operating system: Windows 10 (client), Debian 10 (server)
- Browser: Chrome 88.0.4324.150, Firefox 85
- Build Version (Bottom of the page): 2.18.1

## Additional Context

I would propose a twofold fix: first, and easiest, a CSS fix to create a "placeholder" for the image, so it doesn't change size and move other things around if/when it loads. Second, unicode-friendly icon generation. An example of these type of icons in practice is airtable.com (user databases are named using the first two characters, but emoji work fine).

eliykat:
Fixed by https://github.com/bitwarden/jslib/pull/354

PR #1074, opened 2021-07-08
Add support for emojis in Avatar

Hinton:
## Objective
The current avatar component breaks when it gets an emoji followed by a space. To improve the emoji support I've switched to using unicode aware regex to split the characters. Which seems to be the only way to provide support for compounding unicode. According to https://caniuse.com/mdn-javascript_builtins_regexp_unicode this should have good support.

## Screenshots
image
image
image
image
image

Depends on https://github.com/bitwarden/jslib/pull/426
Resolves https://github.com/bitwarden/web/issues/838


addisonbeck (approved):
🐛

PR #1075, opened 2021-07-08
add Kannada language

tgreer-bw:


eliykat (approved):
Easiest PR reviews I've ever done 🤣

PR #1076, opened 2021-07-09
Revert braintree update

cscharf:
Reverts https://github.com/bitwarden/web/pull/1066

Cloudflare is holding on to a pretty good amount of the `dropin.js` file in cache, which has kept PayPal checkout working for a lot of folks and has made it spotty across the board, but it was observed that for those users getting the latest/newest version of `dropin.js` using a newer version of the Braintree JavaScript Client SDK, the forever spinner is causing issues.

We need to revert this change and either figure something else out or find actual root cause and determine a proper fix later on. This change reverts the update in hopes of getting production working again across the board for PayPal.

Also, I've cut a new `rc` branch anticipating another release of the web vault and have version bumped accordingly.

cc: @sevensixseven

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "bitwarden-web",
-      "version": "2.21.1",
+      "version": "2.20.4",
       "hasInstallScript": true,
       "license": "GPL-3.0",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


kspearrin (approved):

Hinton:
I don't think this will solve the issue. The production instance is using 1.28.0 on my machine, and it displays the infinite spinner. But if I cache bust it (change src tag of `dropin.js` to `dropin.js?cachebust=true`) it loads in 1.30.1, which displays the checkout button. Seems more that cloudflare is serving old files.

PR #1081, opened 2021-07-12
Remove send info card from web vault

MGibson1:
# Overview

Send is now an pretty established product. We don't need this card anymore.


Hinton (changes requested):
Looks like you forgot to remove the component from `OssModule`

Hinton (approved):

PR #1086, opened 2021-07-18
typo

colinfrei:
build:dev:watch requires a development.json file, not development.js

Causes the error "[webpack-cli] Error: Cannot find module './config/development.json'" otherwise.


djsmith85 (approved):
@colinfrei Good catch and thank you for your contribution.

PR #1089, opened 2021-07-20
Feature/use hcaptcha if bot

MGibson1:
# Overview

Implement captcha connector for use across clients in an iframe. Also add captcha to login component.

The usage flow is like this:
1. login with user/pass
  a. server rejects with captcha required
2. login with user/pass/catpcha
  a. login successful
  or
  b. login unsuccessful due to captcha -- try again
  or
  c. login unsuccessful due to two-factor required -- Server will provide captcha bypass
3. login with stored user/pass/captcha bypass + new two factor

# Files Changed
* **login.component**: Add captcha if the auth response from the server demands it
* **captcha.html**: Add mobile responsive headers for mobile client and remove script src to be added by our own script. This is necessary to enable automatic language selection using the `hl=` src paramter.
* **captcha.ts**: parse out the data passed in the params. Also send regular updates as to the height and width of the captcha prompt. I tried out using [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver), but this creates a feedback loop when resizing the iframe. The overall size of the window does not change when completing the image selection portion, so if that dive is hidden, we use the offset height of the visible elements (just the checkbox ~80px)
* **common-webauth/common/webauthn-fallback**: extract b64decode method common to iframe components
* **webauthn**: Use the new data params scheme from the `iframe_compenent` changes in bitwarden/jslib#430
* **styles.css**: style the hcaptcha iframe

# Screenshots

### DIFF
--- a/src/app/accounts/login.component.html
+++ b/src/app/accounts/login.component.html
@@ -36,6 +36,7 @@
                         <label class="form-check-label" for="rememberEmail">{{'rememberEmail' | i18n}}</label>
                     </div>
                     <hr>
+                    <div [hidden]="!showCaptcha()"><iframe id="hcaptcha_iframe" height=80></iframe></div>
                     <div class="d-flex">
                         <button type="submit" class="btn btn-primary btn-block btn-submit" [disabled]="form.loading">
                             <span>
--- a/src/connectors/captcha.ts
+++ b/src/connectors/captcha.ts
@@ -14,6 +14,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
 let parentUrl: string = null;
 let parentOrigin: string = null;
+let callbackUri: string = null;
 let sentSuccess = false;
 
 async function init() {
@@ -48,6 +49,7 @@ async function start() {
         error('Cannot parse data.');
         return;
     }
+    callbackUri = decodedData.callbackUri;
 
     let src = 'https://hcaptcha.com/1/api.js?render=explicit';
 
@@ -77,6 +79,9 @@ async function start() {
 
 function captchaSuccess(response: string) {
     success(response);
+    if (callbackUri) {
+        document.location.replace(callbackUri + '?token=' + encodeURIComponent(response));
+    }
 }
 
 function captchaError() {
--- a/src/connectors/captcha.ts
+++ b/src/connectors/captcha.ts
@@ -13,7 +13,6 @@ document.addEventListener('DOMContentLoaded', () => {
 (window as any).captchaError = captchaError;
 
 let parentUrl: string = null;
-let locale: string = null;
 let parentOrigin: string = null;
 let sentSuccess = false;
 
@@ -31,8 +30,6 @@ async function start() {
         error('No data.');
         return;
     }
-
-    locale = getQsParam('locale')
 
     parentUrl = getQsParam('parent');
     if (!parentUrl) {
@@ -55,8 +52,8 @@ async function start() {
     let src = 'https://hcaptcha.com/1/api.js?render=explicit';
 
     // Set language code
-    if (locale) {
-        src += `&hl=${locale ?? 'en'}`;
+    if (decodedData.locale) {
+        src += `&hl=${decodedData.locale ?? 'en'}`;
     }
 
     const script = document.createElement('script');
@@ -72,7 +69,6 @@ async function start() {
     }
 
     hcaptcha.render('captcha', {
-        languageCode: decodedData.locale,
         sitekey: decodedData.siteKey,
         callback: 'captchaSuccess',
         'error-callback': 'captchaError',
--- a/src/connectors/captcha.ts
+++ b/src/connectors/captcha.ts
@@ -1,4 +1,4 @@
-import { getQsParam } from './common';
+import { b64Decode, getQsParam } from './common';
 
 declare var hcaptcha: any;
 
@@ -13,16 +13,17 @@ document.addEventListener('DOMContentLoaded', () => {
 (window as any).captchaError = captchaError;
 
 let parentUrl: string = null;
+let locale: string = null;
 let parentOrigin: string = null;
 let sentSuccess = false;
 
-function init() {
-    start();
+async function init() {
+    await start();
     onMessage();
-    info('ready');
+    watchHeight();
 }
 
-function start() {
+async function start() {
     sentSuccess = false;
 
     const data = getQsParam('data');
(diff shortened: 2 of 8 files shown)


src/connectors/webauthn.ts
```diff
@@ -1,20 +1,22 @@
...
+import { b64Decode, getQsParam } from './common';
+import { buildDataString, parseWebauthnJson } from './common-webauthn';
 
 // tslint:disable-next-line
 require('./webauthn.scss');
 
 document.addEventListener('DOMContentLoaded', () => {
     init();
 
-    const text = getQsParam('btnText');
```
Hinton, line 10: Just a heads up. If we do change the protocol, we'll need to include support for both variants for a while. To avoid old clients from getting locked out.

src/connectors/webauthn.ts
```diff
@@ -1,20 +1,23 @@
...
-    if (text) {
+    parseParameters();
+    if (btnText) {
```
Hinton, line 11: It feels a bit wierd to use this variable here, when it's declared below.

src/connectors/captcha.ts
```diff
@@ -79,7 +112,24 @@ function success(data: string) {
...
 
-function info(message: string) {
-    parent.postMessage('info|' + message, parentUrl);
+function info(message: string | object) {
+    parent.postMessage('info|' + JSON.stringify(message), parentUrl);
+}
+
+async function watchHeight() {
+    const imagesDiv = document.body.lastChild as HTMLElement;
+    while (true) {
```
Hinton, line 121: Do we need to run this forever?

src/connectors/captcha.ts
```diff
@@ -40,15 +41,47 @@ function start() {
...
+        error('Cannot parse data.');
+        return;
+    }
+    callbackUri = decodedData.callbackUri;
+
+    let src = 'https://hcaptcha.com/1/api.js?render=explicit';
+
+    // Set language code
+    if (decodedData.locale) {
+        src += `&hl=${decodedData.locale ?? 'en'}`;
```
Hinton, line 58: Do we need to process this in any way? Do hcaptcha support all locales.

MGibson1: If hcaptcha can't find the locale we request it uses it's default, which attempts to detect the appropriate locale. The only reason I'm doing this is to try and match what you're already specified to bitwarden.

src/connectors/captcha.ts
MGibson1: we don't know when the captcha will be begun or solved. Are you wanting to stop after some number of seconds? Technically we could stop after a success, but I don't see a huge benefit to that since that _should_ mean the form is about to be submitted

src/app/accounts/login.component.html
```diff
@@ -36,6 +36,7 @@
                         <label class="form-check-label" for="rememberEmail">{{'rememberEmail' | i18n}}</label>
                     </div>
                     <hr>
+                    <div [hidden]="!showCaptcha()"><iframe id="hcaptcha_iframe" height=80></iframe></div>
```
cscharf, line 39: It would be awesome if this was centered similar to the buttons below it (same as the other forms, should be aligned more with the buttons below it since it's a blocky looking element located just above the buttons).

MGibson1: Better?
image

MGibson1: I don't really like either. I wanted it to be the full width but I don't want to mess with the internal styling of the element.

I went with aligned to the other inputs but I can see your logic too.  Maybe we should move above the <hr> and left-align?

danielleflinn: I agree that full width would be ideal; but if that is not an option due to lift, I prefer left aligned.

danielleflinn: Sorry @MGibson1 I missed the note to move the captcha above the line; I agree with this idea as well. It helps organize all the fields the user needs to address in one area with the action buttons below. 
Screen Shot 2021-07-22 at 8 49 28 AM

MGibson1:
mobile captcha view

src/app/accounts/login.component.html
```diff
@@ -30,11 +30,12 @@
                             <a routerLink="/hint">{{'getMasterPasswordHint' | i18n}}</a>
                         </small>
                     </div>
-                    <div class="form-check">
+                    <div class="form-check mb-3">
                         <input type="checkbox" class="form-check-input" id="rememberEmail" name="RememberEmail"
                             [(ngModel)]="rememberEmail">
                         <label class="form-check-label" for="rememberEmail">{{'rememberEmail' | i18n}}</label>
                     </div>
+                    <div class="mb-n3" [hidden]="!showCaptcha()"><iframe id="hcaptcha_iframe" height=80></iframe></div>
```
cscharf, line 38: I know it's likely not required but I would have more warm-and-fuzzies if the `height` attribute value was enclosed in double-quotes.

src/app/services/services.module.ts
```diff
@@ -69,6 +68,7 @@ import { FileUploadService as FileUploadServiceAbstraction }  from 'jslib-common
 import { FolderService as FolderServiceAbstraction } from 'jslib-common/abstractions/folder.service';
 import { I18nService as I18nServiceAbstraction } from 'jslib-common/abstractions/i18n.service';
 import { ImportService as ImportServiceAbstraction } from 'jslib-common/abstractions/import.service';
+import { LogService } from 'jslib-common/abstractions/log.service';
```
cscharf, line 71: Nice catch/move!

cscharf (commented):
I think I'm good with these changes... a couple of questions/comments from @Hinton I would like for him to reevaluate.

src/app/accounts/login.component.html
MGibson1: haha good catch I totally agree

danielleflinn (dismissed):
Latest screenshots in Conversation look good to me.

cscharf (approved):

src/connectors/captcha.ts
Hinton: Gotcha, I didn't notice initially that we had to resize the iframe to fit the click.

Hinton (approved):

cscharf (approved):

cscharf (dismissed):

cscharf (approved):

PR #1091, opened 2021-07-21
[Provider] Add provider link in navbar

Hinton:
## Objective
Adds a link to the users provider. The link is only visible if the user belongs to a provider, and if the user only belongs to a single provider, it will link directly to that provider.

### Code Changes
- **bitwarden_license/src/app/providers/providers-routing.module.ts**: Add url for viewing providers.
- **src/app/layouts/navbar.component.{ts, html}**: Add link to navbar.
- **src/app/providers/providers.component.{ts, html}**: Add support for a more full-screen provider view.

https://app.asana.com/0/1198901840263430/1200623666598577

### DIFF
--- a/src/app/layouts/navbar.component.ts
+++ b/src/app/layouts/navbar.component.ts
@@ -5,7 +5,11 @@ import {
 
 import { MessagingService } from 'jslib-common/abstractions/messaging.service';
 import { PlatformUtilsService } from 'jslib-common/abstractions/platformUtils.service';
+import { SyncService } from 'jslib-common/abstractions/sync.service';
 import { TokenService } from 'jslib-common/abstractions/token.service';
+import { UserService } from 'jslib-common/abstractions/user.service';
+
+import { Provider } from 'jslib-common/models/domain/provider';
 
 @Component({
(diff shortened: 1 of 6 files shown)


src/app/layouts/navbar.component.ts
```diff
@@ -27,6 +32,12 @@ export class NavbarComponent implements OnInit {
         if (this.name == null || this.name.trim() === '') {
             this.name = this.email;
         }
+
+        // Ensure provides are loaded
+        if (await this.syncService.getLastSync() == null) {
+            await this.syncService.fullSync(false);
+        }
+        this.providers = await this.userService.getAllProviders();
```
cscharf, line 40: Could this happen perhaps after init, I know you mentioned a reactive model which would be great, but seems like firing this off asynchronously after paint/render and reactively showing the navigation would be a better approach than blocking the rendering of the component itself on init.

Hinton: We don't block of rendering. `OnInit` does not await the promise before rendering.

cscharf (approved):

PR #1092, opened 2021-07-22
Use the helper methods in EnvironmentService

Hinton:
## Objective
It was difficult to figure out which urls were used through the application, since there were no central place to store/retrieve them. To resolve this, I refactored the `EnvironmentService` to keep the state private, and instead expose helper functions for retrieving the appropriate urls.

### Testing Considerations
We should verify all api calls, links, etc still works as expected in our different environments. As well as ensuring notification service gets initialized, since the logic no longer happens in `environmentService`.

Depends on:
https://github.com/bitwarden/jslib/pull/435

Related to:
https://app.asana.com/0/1169444489336079/1200631805316488


cscharf (commented):
This looks good once jslib has been updated/completed.

cscharf (approved):

PR #1093, opened 2021-07-22
Ensure Angular is running as production

Hinton:
## Objective
In #994, we changed `process.env.ENV` to use `ENV` instead of `NODE_ENV`. This had the side effect of also propagating to other parts where we want to use `NODE_ENV`.

### Code Changes
- **app/main.ts**: Angular now only runs in development mode when NODE_ENV==production.
- **webpack.config.js**: Added `NODE_ENV`, and enforce it to be either `production` or `development`.
- **src/services/webPlatformUtils.service.ts**: Changed `isDev` to use `NODE_ENV` to be more consistent with other platforms. And avoid development settings from being exposed in production.
- **src/app/settings/payment.component.ts**: Ensured `ENV` is still used to toggle paypal sandbox.
- **src/app/settings/add-credit.component.ts**: Ensured `ENV` is still used to toggle paypal sandbox.

https://app.asana.com/0/1169444489336079/1200631708822147

### DIFF
--- a/src/app/settings/add-credit.component.ts
+++ b/src/app/settings/add-credit.component.ts
@@ -47,7 +47,7 @@ export class AddCreditComponent implements OnInit {
 
     constructor(private userService: UserService, private apiService: ApiService,
         private platformUtilsService: PlatformUtilsService) {
-        if (platformUtilsService.isDev()) {
+        if (process.env.ENV !== 'production') {
             this.ppButtonFormAction = WebConstants.paypal.buttonActionSandbox;
             this.ppButtonBusinessId = WebConstants.paypal.businessIdSandbox;
         }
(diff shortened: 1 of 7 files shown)


src/app/settings/add-credit.component.ts
```diff
@@ -47,7 +47,7 @@ export class AddCreditComponent implements OnInit {
 
     constructor(private userService: UserService, private apiService: ApiService,
         private platformUtilsService: PlatformUtilsService) {
-        if (platformUtilsService.isDev()) {
+        if (process.env.ENV !== 'production') {
```
cscharf, line 50: Why use `ENV` for this check vs. `NODE_ENV`? Would it perhaps make more sense to use `ENV || NODE_ENV` for all checks?

cscharf (commented):
Just an open question on the differing use of `ENV` vs. `NODE_ENV` for PayPal

src/app/settings/add-credit.component.ts
Hinton: It's been a topic of discussion between me and @joseph-flinn. I've changed it to `env && node_env`. But feel free to change it if we think it should be something else.

cscharf (approved):

PR #1097, opened 2021-07-24
Set env with EnvironmentPlugin

MGibson1:
# Overview

Fixes QA env issue apparently introduced in #1093 

Switch to using `EnvironmentPlugin` to set environment rather than `DefinePlugin`. It appears that the definition of the `process.env` object isn't settings our default values correctly.

I believe the two environment replacements _should_ be identical, but we're seeing issues in qa testing with missing settings. It looks like webpack is replacing the env variable with a variable rather than a string. This plugin just replaces the entire if statement with a boolean

## DefinePlugin:
image

## EnvironmentPlugin
image

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -137,14 +137,12 @@ const plugins = [
         filename: '[name].[hash].css',
         chunkFilename: '[id].[hash].css',
     }),
-    new webpack.DefinePlugin({
-        'process.env': {
-            'ENV': JSON.stringify(ENV),
-            'NODE_ENV': NODE_ENV === 'production' ? 'production' : 'development',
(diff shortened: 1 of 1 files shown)


Hinton (dismissed):
Looks good! Since `EnvironmentPlugin` runs all the values through `JSON.stringify` we can if we want remove the manual calls to `JSON.stringify`.

MGibson1:
I think we _have_ to remove the manual `stringify`s. after all, `JSON.stringify('string')` ≠ `JSON.stringify(JSON.stringify('string'))`

addisonbeck (approved):

PR #1103, opened 2021-07-29
Use 2fa.directory API v3 in inactive 2FA report

### LINKED ISSUE #893: Inactive 2FA Report - Some websites reports are missing
## Describe the Bug

I found that the number of websites in "Inactive 2FA Report" is less than 1Password (with the same amount of password data), lacking of website such as facebook.com, autodesk.com, jetbrains.com, linkedin.com, teamviewer.com, etc. 
After checking [#839](https://github.com/bitwarden/web/issues/839) , I searched the websites above on 2fa.directory, they are both showed support for software token. I'm so confused, why are most of the results normal, but some missing？I‘m not sure it is only abnormal to me, so I post here. Thanks a lot.

## Steps To Reproduce

1. Login to Web Vault
2. Navigate to Tools Tab
3. Click Inactive 2FA Report

## Expected Result

Inactive 2FA Report might be matching with 2fa.directory. 

## Actual Result

Inactive 2FA Report missing websites such as
- adobe.com
- autodesk.com
- cloudflare.com
- facebook.com
- jetbrains.com
- linkedin.com
- teamviewer.com

## Environment

- Operating system: Windows 10 Pro Build 20H2
- Browser: Vivaldi 3.7.2218.45 (Chrome/89.0.4389.91)
- Build Version (Bottom of the page): 2.19.0

### LINKED ISSUE #982: Inactive 2FA report: Domain rules ignored
## Describe the Bug

When running the inactive 2FA report, domain rules are not applied.

## Steps To Reproduce

1. Create an account on a page which provides 2FA and is listed in domain rules
2. Use domain which is not provided at [2fa.directory](https://2fa.directory/)
3. Run inactive 2FA report

In my case I have an account at [netcup.de](https://netcup.de/). But netcup's login page is at [customercontrolpanel.de](https://www.customercontrolpanel.de/). netcup is listed at domain rules and [2fa.directory](https://2fa.directory/) but doesnt show up at inactive 2FA report.

## Expected Result

The account should get listed at inactive 2FA report.

## Actual Result

Account gets not listed at inactive 2FA report.

## Environment

- Operating system: Windows 10
- Browser: Chrome 90.0.4430.212
- Build Version: 2.20.1

ApCoder123:
https://community.bitwarden.com/t/improve-inactive-2fa-report/32045

Closes #982
Closes #893


djsmith85:
Hi @ApCoder123 Thank you for your contribution. 

I've just tested this and although it does solve the above mentioned issue, it misses other domains which should pop up. I used for example `https://www.avanza.se/` and `netcup.de`. With the original version I would get a hit for `avanza` but with your version it would not show up. If I remove the `http://` from the URI-input it would show up. 

If you are still willing to spend some more time on this, could you please do some more testing and make sure that all entries show up.

Kind Regards,
Daniel

djsmith85 (changes requested):
Hi @ApCoder123 Thank you for your contribution. 

I've just tested this and although it does solve the above mentioned issue, it misses other domains which should pop up. I used for example `https://www.avanza.se/` and `netcup.de`. With the original version I would get a hit for `avanza` but with your version it would not show up. If I remove the `http://` from the URI-input it would show up. 

If you are still willing to spend some more time on this, could you please do some more testing and make sure that all entries show up.

Kind Regards,
Daniel

ApCoder123:
Ok, I've made some changes now which should resolve these issues. Rather than just test the hostname, I have modified it to check the base domain (should close #893 as well), which should resolve these issues. 

I believe the reason for `https://www.avanza.se` not showing up was because when the `url` field was being grabbed from the v2 API, `Utils.getHostname()` would cause `www.avanza.se` to be added to the services map, however when the v3 API provides just the domain, `www.avanza.se` wouldn't match `avanza.se`, causing it not to be shown. The checking of the main domain in the `setCiphers()` function should resolve this now.

ApCoder123:
Hi @djsmith85
Sorry to bother you, but are there any updates on when this might be able to be merged? Thanks

djsmith85 (changes requested):
Hi @ApCoder123 I'm really sorry it has taken me so long to get back to you. 

I've had another look at this and based on the data we are retrieving I think we can get rid of the insertation and comparison with hostnames. If we only include the domain, which is present on every item from 2fa-directory json, we could only check for that in the setCiphers method.

After retrieving the domain from the cipher's uri, we should remove the `www.` (replace).

For the loop in `load2fa` could you please invert the `if's` and use `continue`, to reduce if-nesting.

src/app/tools/inactive-two-factor-report.component.ts
```diff
@@ -53,13 +53,13 @@ export class InactiveTwoFactorReportComponent extends CipherReportComponent impl
                 for (let i = 0; i < c.login.uris.length; i++) {
                     const u = c.login.uris[i];
                     if (u.uri != null && u.uri !== '') {
-                        const hostname = Utils.getHostname(u.uri);
-                        if (hostname != null && this.services.has(hostname)) {
-                            if (this.services.get(hostname) != null) {
-                                docs.set(c.id, this.services.get(hostname));
+                        const uri = u.uri.replace("www", "");
```
djsmith85, line 56: This should be using single-quotes.

src/app/tools/inactive-two-factor-report.component.ts
```diff
@@ -78,26 +78,21 @@ export class InactiveTwoFactorReportComponent extends CipherReportComponent impl
...
-                            const hostname = Utils.getHostname(service.url);
-                            if (hostname != null) {
-                                this.services.set(hostname, service.doc);
-                            }
-                        }
-                    }
+        for (const service of responseJson) {
+            const serviceData = service[1];
+            if (serviceData.domain == null) continue;
+            if (serviceData.documentation == null) continue;
```
djsmith85, line 89: Please add curly braces, for one-liner if-statements

djsmith85 (changes requested):
@ApCoder123 Only requesting some minor styling changes and then this is looking good. Thank you again for being so patient

src/app/tools/inactive-two-factor-report.component.ts
```diff
@@ -53,13 +53,13 @@ export class InactiveTwoFactorReportComponent extends CipherReportComponent impl
                 for (let i = 0; i < c.login.uris.length; i++) {
                     const u = c.login.uris[i];
                     if (u.uri != null && u.uri !== '') {
-                        const hostname = Utils.getHostname(u.uri);
-                        if (hostname != null && this.services.has(hostname)) {
-                            if (this.services.get(hostname) != null) {
-                                docs.set(c.id, this.services.get(hostname));
+                        const uri = u.uri.replace('www', '');
```
djsmith85, line 56: Missing a dot, should be `www.`

src/app/tools/inactive-two-factor-report.component.ts
```diff
@@ -78,26 +78,21 @@ export class InactiveTwoFactorReportComponent extends CipherReportComponent impl
...
-                            const hostname = Utils.getHostname(service.url);
-                            if (hostname != null) {
-                                this.services.set(hostname, service.doc);
-                            }
-                        }
-                    }
+        for (const service of responseJson) {
+            const serviceData = service[1];
+            if (serviceData.domain == null) { continue }
+            if (serviceData.documentation == null) { continue }
```
djsmith85, line 89: Please add line breaks, we don't use single-line if's and also semicolons after the `continue`s

djsmith85 (changes requested):

djsmith85 (approved):
@ApCoder123 This is looking good now. Thank you very much for your time and patience.
Placing a `hold` label as we're gearing up to cut our release-candidate branches this week, will merge after that.

PR #1104, opened 2021-07-29
Show only business org create form if provider

MGibson1:
# Overview

Fixes: https://app.asana.com/0/1153292148278596/1200652187192755/f

Providers can only create business-level organizations. Alter org-plans form accordingly.

Depends on #1101

### DIFF
--- a/bitwarden_license/src/app/providers/clients/add-organization.component.ts
+++ b/bitwarden_license/src/app/providers/clients/add-organization.component.ts
@@ -17,7 +17,10 @@ import { ProviderService } from '../services/provider.service';
 
 import { Organization } from 'jslib-common/models/domain/organization';
 import { Provider } from 'jslib-common/models/domain/provider';
+import { ApiService } from 'jslib-common/abstractions';
+import { PlanType } from 'jslib-common/enums/planType';
 
+const DisallowedPlanTypes = [PlanType.Free, PlanType.FamiliesAnnually2019, PlanType.FamiliesAnnually];
 
 @Component({
     selector: 'provider-add-organization',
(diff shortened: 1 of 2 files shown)


cscharf:
@MGibson1 , is this ready for review?

MGibson1:
It is now :)

bitwarden_license/src/app/providers/clients/add-organization.component.ts
```diff
@@ -48,7 +52,12 @@ export class AddOrganizationComponent implements OnInit {
 
         this.provider = await this.userService.getProvider(this.providerId);
 
-        this.organizations = (await this.userService.getAllOrganizations()).filter(p => p.providerId == null);
+        const candidateOrgs = (await this.userService.getAllOrganizations()).filter(o => o.providerId == null);
+        const allowedOrgsIds = await Promise.all(candidateOrgs.map(o => this.apiService.getOrganization(o.id))).then(orgs =>
```
vincentsalucci, line 56: Just a thought - is this `load` method call behind a spinner since we're potentially making _N_ api calls back to back? Could potentially take some time if there are a lot of `candidateOrgs`

MGibson1: It is. I thought about adding the org plan to the info in userService, but it kind of breaks our paradigm of keeping that model reserved for user-specific data.

in the end I think the separation of concerns is better and if we run into issues with loading times, the better solution is to make a specialized api call for this.

vincentsalucci: Seems good to me!

vincentsalucci (approved):

bitwarden_license/src/app/providers/clients/add-organization.component.ts
```diff
@@ -17,7 +17,10 @@ import { ProviderService } from '../services/provider.service';
 
 import { Organization } from 'jslib-common/models/domain/organization';
 import { Provider } from 'jslib-common/models/domain/provider';
+import { ApiService } from 'jslib-common/abstractions';
```
vincentsalucci, line 20: Just a note - should fix these import groups so Linter doesn't yell(ow) at us when building.

MGibson1: dang you're right. I'll fix on the next

PR #1123, opened 2021-08-12
Do not call parent if callback given

MGibson1:
# Overview

fixes: https://app.asana.com/0/0/1200777695141203/f

Android < 10 was OK with redirecting from captcha following a success message posted to parent. However, newer versions are not. These versions were never posting to the callback uri.


mportune-bw (approved):

MGibson1:
picked to `rc`

PR #1141, opened 2021-08-20
 Fix role not being displayed in organization user table

Hinton:
## Objective
In the provider user refactor, the role column disappeared from the organization manage page. The root cause seems to be a typo which references the `providerUserType` instead of `OrganizationUserType`.

Note: Will be cherry picked to `rc`.


cscharf (approved):

cscharf:
cherry picked to `rc`

PR #1151, opened 2021-08-24
Set urls from config file

MGibson1:
# Overview

Simplifies url settings to a single config file that will be webpacked into constants. These changes allow us to more simply specify a complete url set for production and qa environments. The previous local override file, `development.json` has been replaced by a required `development.json` and an override file of `local.json`.

Other changes:
* Do not provide devServer details to non-development environments. This doesn't actually change anything with deployment since this setting is only used by the webpack server. However, it does help clarify that `devServer` is only useful for development environments
* Remove `watch` scripts for non-development builds. We won't be editing files if we're not in dev.
* `config.js` changes improve print output of deeply nested configurations. Improved missing file error handling. Add `local.json` as an ultimate override file.

>Note: this also fixes the dev issue where the `business portal` button didn't work.


src/app/services/services.module.ts
```diff
@@ -146,15 +146,12 @@ export function initFactory(): Function {
     return async () => {
         await (storageService as HtmlStorageService).init();
 
-        if (process.env.ENV !== 'production' || platformUtilsService.isSelfHost()) {
+        const urls = process.env.URLS as Urls;
+        urls.base ??= window.location.origin;
+        if (platformUtilsService.isSelfHost()) {
```
Hinton, line 151: Do we still need this conditional with the change? Since we get the urls from `webpack` we should be able to reliably set urls to an empty object which would only resolve base.

Hinton (commented):
Minor question.

I also wonder if we perhaps should inject the whole `config` file as a process.env variable. Would allow us to add additional settings in the future should the need arise.

src/app/services/services.module.ts
MGibson1: I didn't want to mess with our builds but I decided it was a good idea.

config.js
```diff
@@ -1,28 +1,30 @@
...
-
-    const baseConfig = require('./config/base.json');
-    const overrideConfig = envOverrides.hasOwnProperty(envName) ? envOverrides[envName]() : {};
-
     return {
-        ...baseConfig,
-        ...overrideConfig
+        ...require('./config/base.json'),
+        ...loadConfig(envName),
+        ...loadConfig('local'),
```
joseph-flinn, line 5:
```suggestion
        ...require('./config/base.json'),
        ...loadConfig(envName),
```

Thoughts on treating `local.json` like any other `envName`? I can see the possibility of someone having a `local.json` file overwriting (I believe order of object expansion has overwrite rules) or adding undesired config values. The process can be outlined/updated in the README.

joseph-flinn (changes requested):
The config section of the README should probably be updated to reflect the new dev flow to help smooth the transition

Hinton:
We also use `process.env.ENV !== production` in `add-credit.component.ts`. Which will either need to be updated to `cloud`, or extract those variables to the config files.

MGibson1:
Thanks, @Hinton, those references are updated.

config.js
MGibson1: It think local should be special in that you can run any config you want and override some values without checking them in to version control.

I don't foresee it being very useful, but sometimes people have wonky setups and it should help with community setup of dev environments

config.js
```diff
@@ -1,28 +1,30 @@
...
+    console.log(JSON.stringify(configObj, null, 2));
     console.log(`${"=".repeat(repeatNum)}`)
 }
 
+function loadConfig(configName) {
+    try {
+        return require(`./config/${configName}.json`);
+    }
+    catch (e) {
+        if (e instanceof Error && e.code === "MODULE_NOT_FOUND")
```
Hinton, line 21: I believe we generally always include brackets in our statements.

Hinton (approved):

config.js
joseph-flinn: Ah, so local isn't a full environment? Just an override file?

MGibson1: that's the intent, yeah

src/app/settings/payment.component.ts
```diff
@@ -67,7 +67,7 @@ export class PaymentComponent implements OnInit {
         this.stripeScript.src = 'https://js.stripe.com/v3/';
         this.stripeScript.async = true;
         this.stripeScript.onload = () => {
-            this.stripe = (window as any).Stripe(process.env.ENV === 'production' && !platformUtilsService.isDev()  ?
```
Helix-DB3, line 70: Why was stripe removed?

Helix-DB3 (commented):
I have a question for review.

src/app/settings/payment.component.ts
Hinton: It wasn't, we just changed the ENV name.

PR #1164, opened 2021-08-28
Add issue template and template chooser

Hinton:
## Objective
Simplify the workflow for submitting issues and finding help links by adding a new issue template which utilizes github issue forms.

## Screenshots
image
image


.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,77 @@
...
+    id: screenshots
+    attributes:
+      label: Screenshots or Videos
+      description: If applicable, add screenshots and/or a short video to help explain your problem.
+  - type: dropdown
+    id: os
+    attributes:
+      label: What operating system are you seeing the problem on?
+      multiple: true
+      options:
```
eliykat, line 48: While I like the multi-select dropdown here, it doesn't let you specify your OS version or (for Linux) your distro. Maybe a single line text field would be better to capture this info.

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,77 @@
...
+        - Android
+        - iOS
+    validations:
+      required: true
+  - type: dropdown
+    id: browsers
+    attributes:
+      label: What browsers are you seeing the problem on?
+      multiple: true
+      options:
```
eliykat, line 61: Same here - we want to know the version, as sometimes older browsers can have compatibility issues, or they're using FF Nightly instead of FF stable etc.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,14 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Feature Requests
+    url: https://community.bitwarden.com/c/feature-requests/
+    about: Request new features using the Community Forums. Please search existing feature requests before making a new one
```
eliykat, line 5:
```suggestion
    about: Request new features using the Community Forums. Please search existing feature requests before making a new one.
```

Sorry, this is petty but it will annoy me if I don't fix it now.

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,77 @@
+name: Bug Report
+description: File a bug report
+title: "[Bug]: "
```
eliykat, line 3: If we only use Github for bug reports, is it useful adding this to the title and using the label?

At most I think just use the label (in case we change our use of Github in the future), and leave it out of the name (which seems duplicative).

eliykat (changes requested):

.github/ISSUE_TEMPLATE/bug.yaml
Hinton: I'm hesitant to convert it to a free text field since we might have a harder time parsing the results. However I've added a new field to specify version which while it's not required should hopefully let users provide more information.

eliykat (approved):
Nice work! Keen to see how this goes in practice.

PR #1166, opened 2021-08-31
Add notifications to development config

Hinton:
## Objective
When running `npm run build:dev:watch` we currently cannot speak with the notification service since it uses websockets. Usually I would solve this by adding `ws: true` to the proxy configuration, but for some reason that crashes webpack on my machine.

Since we recently introduced support for overriding service urls, I instead changed so the application speaks directly to the notification service.


cscharf (approved):
👍🏻

PR #1167, opened 2021-08-31
[Provider] Verify user is owner of organization

Hinton:
## Objective
We had an issue where users didn't have permission to call organizations/:id, which caused an never-ending spinner in the provider portal. To resolve this, I added a check to ensure the user has owner permission of organization, which is required to add an org either way.

Asana: https://app.asana.com/0/1200578627114106/1200880201964177


MGibson1 (approved):
Good catch, reduces API calls too. Sorry for putting this in 🐛.

PR #1175, opened 2021-09-02
[Policy] Reset Password data bugfix

vincentsalucci:
## Code Changes
- **reset-password.component.html**: Changed `formControlName` to `autoEnrollEnabled`
- **reset-password.component.ts**: Changed data key from `autoEnroll` to `autoEnrollEnabled`


Hinton (approved):
👍 Thanks for finding this.

PR #1176, opened 2021-09-03
Disable personal imports if Personal Ownership policy applies

eliykat:
## Objective

The Personal Ownership policy prevents users from creating items in their personal vault. However, at the moment it does not prevent a user from _importing_ items into their personal vault.

## Code changes

Update `import.component` template to:
* disable the import form controls if the policy applies (with callout)
* reject any import form submission if the policy applies

Server changes are to follow.

## Screenshots

Screen Shot 2021-09-03 at 2 04 43 pm


djsmith85 (approved):
Good catch @eliykat

PR #1181, opened 2021-09-08
Possible fix for blocked nav on some devices

mportune-bw:
I have a device that is blocking navigation (per chrome dev tools) on the success callback for reasons unknown.  After comparing with the captcha connector (which works flawlessly), the only difference I can find is that captcha doesn't do anything else after `document.location.replace` for mobile.  I'm not sure if this is the culprit but it can't hurt to try.


Hinton (approved):
I don't really se how it would help, but lets try it.

PR #1184, opened 2021-09-08
Move WebConstants values to environment config files

joseph-flinn:
## Summary

Moving the WebConstants for our payment systems into the configuration pattern we established in https://github.com/bitwarden/web/pull/1151.

## Testing Notes
This feature branch has been tested on the QA environment and works as expected.


config/base.json
```diff
@@ -1,3 +1,9 @@
 {
-    "urls": {}
+    "urls": {},
+    "stripeKey": "pk_test_KPoCfZXu7mznb9uSCPZ2JpTD",
+    "btKey": "sandbox_r72q8jq6_9pnxkwm75f87sdc2",
```
Hinton, line 4: Could we rename this to `braintreeKey`? It's not obvious at first glance what `bt` stands for.

webpack.config.js
```diff
@@ -148,6 +148,9 @@ const plugins = [
         'APPLICATION_VERSION': pjson.version,
         'CACHE_TAG': Math.random().toString(36).substring(7),
         'URLS': envConfig['urls'] ?? {},
+        'STRIPE_KEY': envConfig['stripeKey'] ?? '',
```
Hinton, line 151: It feels like we are going to keep adding variables here. Could we perhaps do something like:

```
'CONFIG': envConfig['config'] ?? {},
```

And then access it using `(process.env.CONFIG as WebConstants).stripeKey`. Perhaps the cast could even be avoided by adding it to `global.d.ts`.

Hinton (commented):
Some quick thoughts.

webpack.config.js
joseph-flinn: Is there a benefit to minimizing the number of env vars to include in this section? With this proposed change, I would be concerned about future reformatting changes to the config files and/or having to check for the existence of the config value before using in the code. I think expanding this section might be a cleaner way of using and tracking the configuration values, but I might be missing something?

Hinton: It helps in minimizing the number of "mappings" which needs to be done. Take paypal as example, in the configs it's defined as  `paypal`, which in `webpack.config.js` is converted to `PAYPAL_CONFIG` and lastly accessed using `process.env.PAYPAL_CONFIG as PaypalConfig`.

Putting all configs under `config` and not in the object root would also provide a separation between values used in webpack for proxies or allowed hosts.

Hinton (approved):

webpack.config.js
MGibson1: defining in `global.d.ts` would be a good way to define a type available to the various configs. I like type definitions to exist for things.

PR #1186, opened 2021-09-09
Add warning about 2FA when changing account email

eliykat:
## Objective

The "Change email" panel in Settings lets the user change their account email, but it will not update the email address used for two-factor authentication (if that method is enabled). This is not intuitive:

>  there's a non-zero risk of someone getting locked out of their account because they don't realize 2FA codes are being sent to an old email address, which they might not have access to.

## Code changes
* if the user has the two-factor email method enabled, show a warning callout in the change-email component.

## Screenshots
Screen Shot 2021-09-09 at 2 03 31 pm

The warning is hidden if the user is not using email 2FA.

### DIFF
--- a/src/app/settings/change-email.component.ts
+++ b/src/app/settings/change-email.component.ts
@@ -36,13 +36,13 @@ export class ChangeEmailComponent implements OnInit {
     async ngOnInit() {
         const twoFactorProviders = await this.apiService.getTwoFactorProviders();
         for (const p of twoFactorProviders.data) {
-            if (p.type == TwoFactorProviderType.Email) {
+            if (p.type === TwoFactorProviderType.Email) {
                 this.showTwoFactorEmailWarning = p.enabled;
                 break;
             }
         }
     }
- 
+
     async submit() {
         const hasEncKey = await this.cryptoService.hasEncKey();
         if (!hasEncKey) {
--- a/src/app/settings/change-email.component.ts
+++ b/src/app/settings/change-email.component.ts
@@ -1,5 +1,6 @@
 import {
     Component,
+    OnInit,
 } from '@angular/core';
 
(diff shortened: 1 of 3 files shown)


src/app/settings/change-email.component.ts
```diff
@@ -13,22 +14,35 @@ import { UserService } from 'jslib-common/abstractions/user.service';
...
 
     formPromise: Promise<any>;
 
     constructor(private apiService: ApiService, private i18nService: I18nService,
         private toasterService: ToasterService, private cryptoService: CryptoService,
         private messagingService: MessagingService, private userService: UserService) { }
 
+    async ngOnInit() {
+        const twoFactorProviders = await this.apiService.getTwoFactorProviders();
+        for (const p of twoFactorProviders.data) {
```
Hinton, line 38: I believe we could simplify this using `some`, something like (not tested):
```js
this.showTwoFactorEmailWarning = twoFactorProviders.data.some(p => p.type === TwoFactorProviderType.Email && p.enabled)
```

src/locales/en/messages.json
```diff
@@ -899,6 +899,9 @@
   "changeEmail": {
     "message": "Change Email"
   },
+  "changeEmailTwoFactorWarning": {
+    "message": "Proceeding will change your account email address. It will not change the email address used for two-factor authentication. You can change this email address in the Two-Step Login settings if required."
```
Hinton, line 903: I believe the last `if required` is unnecessary and possibly confusing.

Hinton (approved):

Hinton:
I wonder if we should display the 2fa email as well. Would give a reminder of which email was used.

eliykat:
2FA settings are gated behind a password re-prompt so I don't think we should surface them here. I think this is enough to prompt the user.

PR #1187, opened 2021-09-09
Remove redundant error messaging

mportune-bw:
Remove the "WebAuth Error" prefix from WebAuthn error strings

### DIFF
--- a/src/connectors/webauthn.ts
+++ b/src/connectors/webauthn.ts
@@ -119,7 +119,7 @@ function executeWebAuthn() {
 
     navigator.credentials.get({ publicKey: obj })
         .then(success)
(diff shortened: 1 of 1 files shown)


src/connectors/webauthn.ts
```diff
@@ -119,7 +119,7 @@ function executeWebAuthn() {
 
     navigator.credentials.get({ publicKey: obj })
         .then(success)
-        .catch(err => error('WebAuth Error: ' + err));
+        .catch(err => error(err));
```
Hinton, line 122: We could just write

```suggestion
        .catch(error);
```

Hinton (approved):

PR #1190, opened 2021-09-11
Renamed application and removed Bitwarden logos

robartsd:


robartsd:
Accidentally targeted upstream instead of my fork.

PR #1193, opened 2021-09-16
Organization autoscaling

MGibson1:
# Overview

Asana Ticket: https://app.asana.com/0/1200804338582616/1200793427058426/f

Redesign of organization subscription page. Changes seat count to a single field and adds options for controlling seat auto scaling.

# Files Changed
* **clients.components.ts/setup.component.ts/**: linter fixes
* **adjust-seats => adjust-subscription**: this is a rename that includes added functionality. Rather than splitting this component into two (add/remove) it presents a number of seats to target and autoscaling options.
* **organization-subscription**: update design and use new subscription component.
* **oss.module**: register new subscription component
* **messages.json**: messages for new subscription model.

# Screenshots

## Full subscription page
image

## Manage subscription requires max if limit subscription
image

## Manage subscription Max seats set
image

## Manage subscription Max seats reached
image

# Draft Reason

Depends on bitwarden/jslib#487


Hinton:
I think we can remove `(optional)` from the checkboxes. I think the checkbox conveys that it's an optional feature which can be enabled. And once the input is visible, the field is required.

src/app/organizations/settings/organization-subscription.component.ts
```diff
@@ -213,4 +214,14 @@ export class OrganizationSubscriptionComponent implements OnInit {
         return (this.sub.planType !== PlanType.Free && this.subscription == null) ||
             (this.subscription != null && !this.subscription.cancelled);
     }
+
+    get subscriptionDesc() {
+        if (this.sub.maxAutoscaleSeats == this.sub.seats && this.sub.seats != null) {
+            return this.i18nService.t('subscriptionMaxReached', this.sub.seats.toString());
+        } else if (this.sub.maxAutoscaleSeats == null) {
```
Hinton, line 221: `else if` doesn't do anything here. I think we just replace it with `if`. And remove the `else` below as well.

src/app/organizations/settings/adjust-subscription.component.ts
```diff
@@ -0,0 +1,86 @@
...
+    @Input() interval = 'year';
+    @Output() onAdjusted = new EventEmitter();
+
+    formPromise: Promise<any>;
+    limitSubscription: boolean;
+    newSeatCount: number;
+    newMaxSeats: number;
+
+    constructor(private apiService: ApiService, private i18nService: I18nService,
+        private toasterService: ToasterService, private router: Router) { }
```
Hinton, line 34: Router isn't used.

src/app/organizations/settings/adjust-subscription.component.ts
```diff
@@ -0,0 +1,86 @@
...
+            const seatAdjustment = this.newSeatCount - this.currentSeatCount;
+            const request = new OrganizationSubscriptionUpdateRequest(seatAdjustment, this.newMaxSeats);
+            this.formPromise = this.apiService.postOrganizationUpdateSubscription(this.organizationId, request);
+
+            await this.formPromise;
+
+            const messages: string[] = [];
+
+            if (seatAdjustment != 0) {
+                messages.push(this.i18nService.t('adjustedSeats', seatAdjustment.toString()));
```
Hinton, line 53: It feels a bit fragile to compose messages this way.

src/app/organizations/settings/organization-subscription.component.html
```diff
@@ -23,110 +23,60 @@ <h1>
...
+                        </dd>
+                        <dt>{{'nextCharge' | i18n}}</dt>
+                        <dd>{{nextInvoice ? ((nextInvoice.date | date: 'mediumDate') + ', ' + (nextInvoice.amount |
+                            currency:'$'))
+                            : '-'}}
+                        </dd>
+                        </ng-container>
+                        </dl>
+                        </div>
+                        <div class="col-8" *ngIf="subscription">
```
Hinton, line 48: We've already checked `*ngIf=subscription` in the parent container.

src/app/organizations/settings/organization-subscription.component.html
```diff
@@ -23,110 +23,60 @@ <h1>
...
+                                {{i.name}} {{i.quantity > 1 ? '&times;' + i.quantity : ''}} @ {{i.amount |
+                                currency:'$'}}
+                                </td>
+                                <td>
+                                    {{(i.quantity * i.amount) | currency:'$'}} /{{i.interval | i18n}}
+                                </td>
+                                </tr>
+                                </tbody>
+                                </table>
+                                </div>
```
Hinton, line 63: Indention is a bit messed up.

Hinton (changes requested):

src/app/organizations/settings/organization-subscription.component.ts
MGibson1: This `else if` is hit if `maxAutoscaleSeats == null` and `seats ≠ null`. It's if your organization has a seat limit, but you haven't defined a max seat limit.

These are the three possibilities:
1. **`maxAutoscaleSeats == seats && != null**: We've reached the maximum limit of seats, Indicate we can no longer expand
2. **maxAutoscaleSeats == null**: Indicate we will expand indefinitely
3. **maxAutoscaleSeats != null && != seats**: Indicate we will expand until we hit max seats.

src/app/organizations/settings/adjust-subscription.component.ts
MGibson1: Changed this out to using different messages per option. It's a little wonky using 3 bools for an if/else, but I think has a better guarantee of message continuity.

src/app/organizations/settings/organization-subscription.component.ts
Hinton: Sorry for being vague, I meant that using `else if` and `else` in statements that returns values are unnecessary. And that we can just write

```typescript
if (this.sub.maxAutoscaleSeats == this.sub.seats && this.sub.seats != null) {
	return this.i18nService.t('subscriptionMaxReached', this.sub.seats.toString());
}

if (this.sub.maxAutoscaleSeats == null) {
    return this.i18nService.t('subscriptionUserSeatsUnlimitedAutoscale');
}

return this.i18nService.t('subscriptionUserSeatsLimitedAutoscale', this.sub.maxAutoscaleSeats.toString());
```

Hinton (dismissed):
I think it's more obvious. Bit unsure if we actually need to include all information in the notification though.

I think a simple `Seats adjusted` would be sufficient. If not we could have two notifications and simplify the logic.

MGibson1:
I suppose it is silly to just parrot back what you updated from a form submission

Hinton (approved):

PR #1198, opened 2021-09-21
Storybook Experiment

Hinton:
## Objective
This is an experiment showcasing how we can add Storybook to our projects.

It's slightly buggy due to some circular dependencies which needs to be resolved. However other than that it seems to be working. In the long term

I still suspect we should look into moving over to the native angular build pipeline. The Storybook work-around seems a bit fragile and there are differences in how webpack bundles the content which sometimes requires some work-around to get Storybook working. Like not being able to use the `BroadcasterService` from `jslib-angular`.

I had to re-work how we handle dependency injection.

### Screenshots
image

Depends on: https://github.com/bitwarden/jslib/pull/495


storybook/src/stories/Avatar.stories.ts
```diff
@@ -0,0 +1,54 @@
...
+    return new WebCryptoFunctionService(window, platformUtilsService);
+}
+
+export default {
+    title: 'Web/Avatar',
+    component: AvatarComponent,
+    decorators: [
+        moduleMetadata({
+            providers: [
+                { provide: CryptoFunctionServiceAbstraction, useFactory: cryptoFunctionFactory},
```
cscharf, line 32:
```suggestion
                { provide: CryptoFunctionServiceAbstraction, useFactory: cryptoFunctionFactory },
```

Hinton:
Closing this since we decided to not move forward at this stage. The services refactor was moved to https://github.com/bitwarden/web/pull/1313

PR #1206, opened 2021-09-27
[Linked fields] Add Linked Field as custom field type

eliykat:
## Objective

Web implementation of https://github.com/bitwarden/browser/pull/1963.

## Screenshots

Note that the other clients have a "view" page, which displays the `fa-link` icon next to the custom field value to indicate that it's a linked field and not a literal value. I did this because once you add a custom field, there's no explicit marker to indicate what type it is.

However, web doesn't have a view page, and the `fa-link` icon doesn't really fit anywhere in the edit page. I tried to add it into the dropdown select, but that is [not supported on all platforms](https://github.com/bitwarden/browser/pull/1963). So we're basically relying on the dropdown to indicate that it's a linked field and not a literal value. I don't think that's the end of the world, but thought I'd mention it in case anyone had a better idea.

Screen Shot 2021-09-27 at 2 33 58 pm


src/app/vault/add-edit-custom-fields.component.html
```diff
@@ -49,7 +51,16 @@ <h3 class="mt-4">{{'customFields' | i18n}}</h3>
...
                         </div>
                     </div>
+                    <!-- Linked -->
+                    <div class="input-group" *ngIf="f.type === fieldType.Linked">
+                        <select id="fieldValue{{i}}" name="Field.Value{{i}}" class="form-control" [(ngModel)]="f.linkedId"
+                            *ngIf="f.type === fieldType.Linked && cipher.linkedFieldOptions != null"
+                            [disabled]="cipher.isDeleted || viewOnly">
+                            <option *ngFor="let o of linkedFieldOptions" [ngValue]="o.value">{{o.name}}</option>
+                        </select>
+                    </div>
```
MGibson1, line 61: I suggest adding a disabled input or a span with `type=password` and our typical eye to show the value, which would be filled in with the linked value.

For each of these custom field types, the `value` box is the indicator of the custom field type, it should be here too.

MGibson1 (changes requested):

src/app/vault/add-edit-custom-fields.component.html
MGibson1: n/m this is a bad idea from me not thinking through the layout. sorry

MGibson1 (approved):

PR #1208, opened 2021-09-29
Update local web development instructions

MGibson1:
# Overview

1. Update instructions to point a local web instance at production servers.
2. Use `NODE_ENV` to more accurately determine if an instance is running in production or development. The `ENV` setting is more accurately used to indicate the configuration of the services, where `NODE_ENV` is used to indicate a mode of operation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,14 +37,28 @@ npm run build:oss:watch
 
 You can now access the web vault in your browser at `https://localhost:8080`.
 
-If you want to point the development web vault to the production APIs, you can run using:
+If you want to point the development web vault to the production APIs, first create a local configuration file at `./config/local.json` with the following contents:
+
+```json
+{
+    "proxyApi": "https://api.bitwarden.com",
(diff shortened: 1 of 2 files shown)


MGibson1:
Thanks for the help figuring this out, @fredrikburmester

Hinton:
@MGibson1 I added proxy configuration for cloud and qa to simplify switching between environments without having to modify `local.json`.

config.js
```diff
@@ -3,6 +3,11 @@ function load(envName) {
         ...require('./config/base.json'),
         ...loadConfig(envName),
         ...loadConfig('local'),
+        dev: {
```
MGibson1, line 6: @Hinton 
Seems weird we need to load these twice to get a property.

Does the override not work properly?

Hinton: I wanted to support merging `dev` properties, as a common use case is to override `allowedHosts` while keeping the environment specific `proxy`. I suppose we could do a deep merge instead.

PR #1211, opened 2021-09-30
Feature/split manage collections permission

MGibson1:
# Overview

Asana: https://app.asana.com/0/0/1201089124980372/f

Split Manage Collections permissions into Create, Edit, Delete.

## Goal

Splits up collections permissions into *ManageAllCollections* => *CreateNewCollections, EditAnyCollection, DeleteAnyCollection* and *ManageAssignedCollections*=> *EditAssignedCollections, DeleteAssignedCollections*.

The new collections will be presented as *sub*premissions under their previous names. To do this, I created a new visual component, `nested-checkbox`, which creates a parent checkbox and child checkboxes. Selecting the parent checkbox selects all children, selecting a subset of child checkboxes results in the parent being in a indeterminate state, but not checked state. The styling for indeterminate is browser-determined, the screenshots below are from firefox.

Also updates all organization collection permission checks. In general, view checks are updated to a `canView{Assigned,All}Collections` helper, deletes are handled by `Delete{Assigned,Any}Collections`, all else (cipher edits, user changes, collections edits) are handled by `Edit{Assigned,Any}Collections`.

# Files Changed
* **nested-checkbox.component**: Displays a checkbox with an inset set of children
* **organization-layout/collections/manage/add-edit/attachments/ciphers/collections/groupings/vault/bulk-delete**: Permissions update
* **user-add-edit**: Support old *Manage* permissions by detecting null new permissions and falling back to old display. This really shouldn't happen as server and web are upgraded together. However, we do need fallback support for CLI so it made sense to put it in here just in case too
* **oss-routing.module/organization-type-guard**: Use new permissions
* **oss.module**: use new nested-checkbox
* **messages.json**: Checkbox labels for new permissions
* **forms.scss**: indent nested-checkbox children

# Draft Reason

- [x] Need to rebase onto master
- [x] Depends on https://github.com/bitwarden/jslib/pull/504

### DIFF
--- a/src/app/organizations/manage/user-add-edit.component.html
+++ b/src/app/organizations/manage/user-add-edit.component.html
@@ -93,9 +93,9 @@
                                     </div>
                                 </div>
                                 <!--  -->
-                                <app-dependent-checkbox *ngIf="!fallbackToManageAssignedCollections" parentId="manageAssignedCollections"
+                                <app-nested-checkbox *ngIf="!fallbackToManageAssignedCollections" parentId="manageAssignedCollections"
                                     [checkboxes]="manageAssignedCollectionsCheckboxes">
-                                </app-dependent-checkbox>
+                                </app-nested-checkbox>
                             </div>
                         </div>
                         <div class="col-6">
@@ -149,9 +149,9 @@
                                     </div>
                                 </div>
                                 <!--  -->
-                                <app-dependent-checkbox *ngIf="!fallbackToManageAllCollections" parentId="manageAllCollections"
+                                <app-nested-checkbox *ngIf="!fallbackToManageAllCollections" parentId="manageAllCollections"
                                     [checkboxes]="manageAllCollectionsCheckboxes">
-                                </app-dependent-checkbox>
+                                </app-nested-checkbox>
                                 <div class="form-group mb-0">
                                     <div class="form-check mt-1 form-check-block">
                                         <input class="form-check-input" type="checkbox" name="manageGroups"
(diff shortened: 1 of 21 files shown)


src/app/organizations/vault/add-edit.component.ts
```diff
@@ -55,14 +55,14 @@ export class AddEditComponent extends BaseAddEditComponent {
     }
 
     protected loadCollections() {
-        if (!this.organization.canManageAllCollections) {
+        if (!this.organization.canEditAnyCollection) {
```
Hinton, line 58: I know this wasn't really changed in your code, but would it be possible to just check if `this.collections` is set and if so just return it? Since we only set it when `canEditAnyCollection == true`.

Hinton (approved):
I think it looks good, it's confusing with all the sprinkled permissions checks in multiple levels since it's not obvious which flow everything has.

src/app/organizations/manage/user-add-edit.component.html
```diff
@@ -80,7 +80,8 @@ <h3 class="mt-4 d-flex">
                             <div class="mb-3">
                                 <label class="font-weight-bold mb-0">Manager Permissions</label>
                                 <hr class="my-0 mr-2" />
-                                <div class="form-group mb-0">
+                                <!-- Deprecated Sep 29 2021 -->
```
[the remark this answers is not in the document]
MGibson1: we do now 🥳

src/app/organizations/vault/add-edit.component.ts
MGibson1: This is a confusing bit of logic, but we can't do that without some pretty big refactors.

What's going on is this:
1. `collections` defaults to an empty list
2. the organization vault component has extra collections context and so takes the opportunity to set `collections` to all collections if it can.
3. the organization add-edit component sets its own collections otherwise.

PR #1212, opened 2021-10-01
Various Dark Theme fixes per QA feedback

eliykat:
## Objective

Address QA feedback on the new Dark Theme.

The defects and all screenshots are listed here: https://app.asana.com/0/1200578627114106/1171263614208497/f.

## Code changes
* `index.html` and `theme.js`
  * move the javascript that sets the initial theme out of `index.html` and into a dedicated `theme.js`. This should solve CORS issues with using inline javascript. 
  * Webpack config has been updated to use `file-loader` for `theme.js`, to avoid it being bundled.
  * make initial theme setting logic more robust - if the theme class is not set on the `html` element (for whatever reason), then the entire app looks broken which is fairly catastrophic. To make sure that theming is always applied no matter what, I've added the `theme_light` class in the html itself as the default theme, and `theme.js` will only change it if it detects that the user has saved a different theme.
* Dark theme fixes:
  * fix theming on event log table, billing table, invoice list
  * fix toast text colours (I also restructured the toast scss while I was at it)
  * fix the red danger text in dropdown menus
  * fix background-color on hover over table rows and dropdown menus
  * apply theming to the little icons that browsers put in input fields, e.g. the calendar icon that appears in the date picker
* Light theme fixes:
  * fix navbar colour on self-hosted instances (should use the darker blue, not our brand blue)
* Changes that affect both themes:
  * use the callout component for the "Disable Send" warning - otherwise it's too bright in dark mode. I discussed with Danielle and we agreed this was an issue of the wrong layout being used rather than the colours themselves.
  * remove unneeded theming when hovering over links
  * fix the styling on the `enterprise2` registration layout - basically just needed to revert all changes to this seeing as how we're not doing any special theming on it for the moment.


src/index.html
```diff
@@ -8,22 +8,8 @@
...
-                theme = 'dark';
-            }
-            else {
-                theme = 'light';
-            }
-            document.documentElement.classList.add('theme_' + theme);
-        })();
-    </script>
+    <script src="scripts/theme.js" crossorigin="anonymous" type="application/javascript"
+        integrity="lmiHgZz6uiAz29i3Q4cALy8IopsEBVAk/6LBcYFMHaDKwA/Cyfod8m5wuqSCjrYL"></script>
```
dltmurphy, line 12: The integrity attribute needs the SHA version as a prefix so it would be `integrity="sha512-lmiHgZz6..."` or whichever SHA you used to generate the hash

[Using Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#using_subresource_integrity)

eliykat: Thanks, this should be fixed now

src/index.html
```diff
@@ -8,22 +8,8 @@
...
-            } else if (theme?.indexOf('dark') > -1) {
-                theme = 'dark';
-            }
-            else {
-                theme = 'light';
-            }
-            document.documentElement.classList.add('theme_' + theme);
-        })();
-    </script>
+    <script src="scripts/theme.js" crossorigin="anonymous" type="application/javascript"
```
Hinton, line 11: We should add the cache buster.

Hinton (changes requested):

src/index.html
Hinton: We could also replace this logic with https://github.com/architgarg/html-webpack-injector but it might overcompensate things.

src/scss/pages.scss
```diff
@@ -160,124 +160,65 @@ app-user-billing {
...
+                position: absolute;
+                top: -76px;
+                transform: skewY(-3deg);
+                width: 100%;
+                z-index: -1;
+            }
+            img.logo {
+                margin: 12px 0 0;
+                max-width: 284px;
+                height: 57px;
```
dltmurphy, line 184: Can you alphabetise this to match the other elements?

src/scss/base.scss
```diff
@@ -177,6 +174,34 @@ code {
     }
 }
 
+.bg-success {
```
dltmurphy, line 177: And alphabetise these too?

src/index.html
eliykat: I wasn't _really_ happy with how I'd integrated this with our build process, so I revisited bundling it using Webpack. In some ways it makes it more complex, but I think it's better to have everything in the same build process. Here's what I've done:

* added `babel-loader` and dependencies to `devDependencies`. These are the latest versions.
* use `babel-loader` instead of `file-loader` to process `theme.js`
* move `theme.js` out of the `scripts` dir so it doesn't get caught by the `CopyWebpackPlugin` which is pointing at this dir
* use `html-webpack-injector` to move `theme.js` into the `<head>` tags. This is a very small and not particularly active repo, but it does one thing and does it in a fairly transparent manner. The alternative would be to manually place our `<script>` tags in the template, which is doable but messier and harder to maintain. Because this is a less reputable source, I've locked it to the current package version (no `~` or `^` in the version number).
* we don't have `integrity` hashes on our webpack bundles, so this is no longer an issue (can be revisited later if desired)
* webpack handles cache busting for us

Webpack is still a learning curve for me, so please tell me if this can be improved. Also let me know if you have any concerns about introducing these dependencies.

src/scss/pages.scss
eliykat: This is all done per the changes you sent me

webpack.config.js
```diff
@@ -48,6 +49,17 @@ const moduleRules = [
             },
         }],
     },
+    {
+        test: /theme\.js$/i,
+        use: [{
+            loader: 'babel-loader',
+            options: {
+                presets: [
+                    ['@babel/preset-env', { targets: "defaults" }]
```
Hinton, line 58: We should use `.browserslistrc` instead.

webpack.config.js
```diff
@@ -48,6 +49,17 @@ const moduleRules = [
             },
         }],
     },
+    {
+        test: /theme\.js$/i,
+        use: [{
+            loader: 'babel-loader',
```
Hinton, line 55: Babel is a fairly large package, I'm not explicitly against adding it, but if we do add it we should also look into adding it to the other rules. It's also a topic worth a discussion whether we should add it or not.

Hinton (approved):

PR #1216, opened 2021-10-05
Remove class attribute from instructions link in 2fa report

djsmith85:
## Objective
Typically we use badges in our reports, to indicate state or more information. In case of the Inactive 2FA report the badge contains a link to the instructions on how to activate 2FA for the specific site. Due to recent changes with the Dark Theme, this has become hardly readable. As to clearly mark it as a link and not a badge looking like a button, I removed the styling.

## Code Changes
- **inactive-two-factor-report.component.html**:  Removed class attribute from instructions link

## Screenshots
**Before:**
with_badge

**After:**
without_badge

### DIFF
--- a/src/app/tools/inactive-two-factor-report.component.html
+++ b/src/app/tools/inactive-two-factor-report.component.html
@@ -41,8 +41,7 @@
                         <small>{{c.subTitle}}</small>
                     </td>
                     <td class="text-right">
-                        <a class="badge badge-primary" href="{{cipherDocs.get(c.id)}}" target="_blank" rel="noopener"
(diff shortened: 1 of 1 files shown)


dltmurphy:
I've got a fix for this missing styling incoming shortly with @eliykat 😊

eliykat:
@djsmith85 this is a defect in the Dark Theme PR. It wasn't intended to affect any light theming, however because web didn't have any theming support at all, we had to restructure the whole scss. That's a bit of a headache for regressions, but we should fix them in the scss, not by changing our templates.

I'll add it as a defect in Asana and submit a separate PR to fix it.

PR #1228, opened 2021-10-11
Add user uses to new permission model

MGibson1:
# Overview

Asana ticket: https://app.asana.com/0/1153292148278596/1201168652396104/f

Permissions on new user add is null, which was reverting our permission model for collections. Server and web are updated in concert, so there should be no issues with versioning issues.


MGibson1:
Yes, 🍒 ⛏️

MGibson1:
picked to `rc`

PR #1229, opened 2021-10-11
[Refactor] Use rxjs first instead of unsubscribe from queryParams

eliykat:
## Objective

As discussed in https://github.com/bitwarden/jslib/pull/470, we currently call `unsubscribe()` on the queryParams object from within its callback. This is a bit of an anti-pattern and can throw an error about the subscription variable being undefined.

## Code changes

Use the rxjs `first` operator instead. This will automatically unsubscribe from the observable after the first value is emitted. We should use this pattern from now on.

This does not strictly depend on https://github.com/bitwarden/jslib/pull/516, but we should include the jslib update in this PR so that it's all together.


addisonbeck:
Merge conflicts RIP 😔

PR #1232, opened 2021-10-12
Compress images u2fkey/yubikey using avif/webp

simon04:
Use modern image formats avif/webp.

Reference: https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types and https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture

Tool used: https://squoosh.app/


simon04:
Sorry for the long silence. Today, I've restored the original JPEG files and improved the quality of the AVIF files.

Hinton:
@simon04 I've rebased this with the prettier changes. added alternatives for u2fkey-mobile and tweaked the avif and webp quality slightly. I believe this is ready to be merged now. Since I've pushed commits to this PR I will get a second reviewer to verify everything looks good.

Thanks for the contribution!

djsmith85 (approved):
Looking good

PR #1274, opened 2021-11-04
Update payment info

justindbaur:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Add additional copy to make it clear that the user will first start a trial and won't be charged until after the trial.
Also make it so the correct copy is shown even when the initial product choice is made because of the query parameter.

Asana Task: https://app.asana.com/0/1153292148278596/1201319764990625/f

## Code changes

* `src/app/layouts/organization-layout.component.html` Added conditional text to be displayed when the user is starting a free trial
* `src/app/settings/organization-plans.component.ts` Gave the `freeTrial` a getter so that it would be checked against the current plan no matter what.
* `src/locales/en/messages.json` Added message for freeTrial copy

## Screenshots

Screen Shot 2021-11-05 at 1 21 33 PM
Screen Shot 2021-11-05 at 1 21 55 PM

## Before you submit
- [x] This change has particular **deployment requirements** (notify the DevOps team)


src/app/settings/organization-plans.component.html
```diff
@@ -219,6 +219,7 @@ <h2 class="spaced-header">{{'summary' | i18n}}</h2>
         <hr class="my-3">
         <h2 class="spaced-header mb-4">{{ (createOrganization ? 'paymentInformation' : 'billingInformation') | i18n}}
         </h2>
+        <p class="text-success" *ngIf="freeTrial && createOrganization">{{ 'freeTrial' | i18n }}</p>
```
cscharf, line 222: Was the verbiage for this text directed from the marketing or product team? Just curious because it's repeated again just below the card information in the `'paymentChargedWithTrial'` string key, which is similar but different, just seems a little redundant, or perhaps like we can at least repeat the same message in both places, or move it to the top (and would no longer need the new string/key)...

justindbaur: The text I added was stolen directly from the Asana task. One of the images in the task also had different text for the bottom text. I did not copy that one because I don't know if they were aware what exactly it said because of the other bug where it wouldn't show the correct text when routed there. 

I agree that is seems a little redundant. I think that maybe just fixing the routing bug where the correct free trial text will be shown when routed to the page and maybe making the text a tad larger would correct the issue.

src/app/settings/organization-plans.component.html
```diff
@@ -234,14 +241,14 @@ <h2 class="spaced-header mb-4">{{ (createOrganization ? 'paymentInformation' : '
             <p class="text-lg"><strong>{{'total' | i18n}}:</strong>
                 {{total | currency:'USD $'}}/{{selectedPlanInterval | i18n}}</p>
         </div>
-        <small class="text-muted font-italic" *ngIf="freeTrial && createOrganization; else paymentChargedImmediately">
+        <!-- <small class="text-muted font-italic" *ngIf="freeTrial && createOrganization; else paymentChargedImmediately">
```
cscharf, line 244: Please remove this block (we don't generally keep commented out code)

cscharf (commented):
one last minor changer

src/app/settings/organization-plans.component.html
justindbaur: Yep, didn't mean to leave that in there.

cscharf (approved):
Awesome, thank you

PR #1277, opened 2021-11-04
Sort Ciphers function implementation

fredrikburmester:
## Type of change
- [x] New feature development

## Objective
This feature adds sorting of the ciphers. It is possible to sort by `name`, `owner` (organization), and `lastEdited` by selecting the desired sorting method or clicking the header `Name`. Reverse sorting is also supported for all methods by pressing the arrow or clicking on the header again. Sorting by `dateCreated` is not yet implemented. 

The addition of this feature has been frequently asked for on the community forum. 

## Code changes
* **src/app/vault/ciphers.component.html:** a new column has been added with a sorting option in the header and the respective sorting information in the column. A select all button has also been added above the name of the cipher in the header of the table.

* **src/app/vault/ciphers.component.ts:** the functionality for the sorting has been added. I've tried to write as good typescript as I can but of course this might need improvement. Here, the sort by `dateCreated` has not been implemented yet. 

* **src/app/vault/bulk-actions.component.html:** the select all option has been removed since it now resides in the table. 

* **src/locales/en/messages.json:** added three english translations for the sorting methods. 

* **src/scss/tables.scss:** some specific CSS has been added to match the provided Figma design. 

## Screenshots
Screenshot 2021-11-04 at 23 40 35

Screenshot 2021-11-04 at 23 40 43

## Testing requirements
- All sorting methods needs to be tested. Both normal order and reverse. 
- Ciphers with empty or no data might need to be tested. 
- Testing with a user that has several orginizations and ciphers would be good.

## Before you submit
(one file I haven't touched has a linting problem. I'm letting it be for now. 
`src/app/organizations/settings/organization-subscription.component.ts:219:40
ERROR: 219:40`

Any comments on code and the feature is very much appreciated, and any help as well. I'm not a professional at all but I really like to contribute. Hope this helps. Thank you.


src/app/vault/bulk-actions.component.html
```diff
@@ -20,15 +20,15 @@
             <i class="fa fa-fw fa-trash-o" aria-hidden="true"></i>
             {{(deleted ? 'permanentlyDeleteSelected' : 'deleteSelected') | i18n}}
         </button>
-        <div class="dropdown-divider"></div>
+        <!-- <div class="dropdown-divider"></div>
```
djsmith85, line 23: This should be removed completely

src/app/vault/ciphers.component.html
```diff
@@ -1,6 +1,35 @@
...
+                    <div (click)="reverseSort()" *ngIf="sortedDescending; else elseBlock" class="input-group-prepend no-bg">
+                        <label  class="input-group-text bg-transparent border-0" for="inputGroupSelect01">↓</label>
+                    </div>
+                    <ng-template #elseBlock>
+                        <div (click)="reverseSort()" class="input-group-prepend">
+                            <label  class="input-group-text bg-transparent border-0" for="inputGroupSelect01">↑</label>
+                        </div>
+                    </ng-template>
+                    <select #option (change)="sortCiphers(option.value)" class="border-0 custom-select bg-transparent font-weight-bold" id="inputGroupSelect01">
+                        <option value="" selected="selected">Sort By</option>
```
djsmith85, line 22: Sort By should also be translatable (i18n)

src/app/vault/ciphers.component.ts
```diff
@@ -44,6 +44,11 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
     actionPromise: Promise<any>;
     userHasPremiumAccess = false;
 
+    organizationNames = new Map<string, string>();
+    sortedDescending = true;
+    sortBy = '' as string;
```
djsmith85, line 49: Instead of `as` use `: type`
```suggestion
    sortBy: string = '';
```

src/app/vault/ciphers.component.ts
```diff
@@ -44,6 +44,11 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+    sortByOptions = ['name', 'owner', 'dateCreated', 'lastEdited'] as (string)[];
```
djsmith85, line 50: Same here
added readonly as these values are fixed on initialization
```suggestion
    readonly sortByOptions: string[] = ['name', 'owner', 'dateCreated', 'lastEdited'];
```

src/app/vault/ciphers.component.ts
```diff
@@ -58,6 +63,7 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
 
     async ngOnInit() {
         this.userHasPremiumAccess = await this.userService.canAccessPremium();
+        this.organizationNames = await this.loadOranizations();
```
djsmith85, line 66: Typo in method name loadOr**g**anizations() and possibly rename to `loadOrganizationNames`

src/app/vault/ciphers.component.ts
```diff
@@ -80,6 +86,90 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+        }
+
+        return organizationNames;
+    }
+
+    getOrganizationById(c: CipherView): string {
+        if (c.organizationId != null) {
+            return this.organizationNames.get(c.organizationId);
+        }
+        return 'Personal';
```
djsmith85, line 106: This should be translatable, maybe use 'myVault'
```suggestion
        return this.i18nService.t('myVault');
```

src/app/vault/ciphers.component.ts
```diff
@@ -80,6 +86,90 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+        if (organizations.length > 0) {
+            organizations.forEach(organization => {
+                organizationNames.set(organization.id, organization.name);
+            });
+        }
+
+        return organizationNames;
+    }
+
+    getOrganizationById(c: CipherView): string {
```
djsmith85, line 102: Possibly rename to getOrganizationNameById and do we need to pass in the whole cipher? If we only pass in the orgId the name would also fit better.

src/app/vault/ciphers.component.ts
```diff
@@ -80,6 +86,90 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+        if (c.organizationId != null) {
+            return this.organizationNames.get(c.organizationId);
+        }
+        return 'Personal';
+    }
+
+    sortCiphers(sortBy: string): void {
+        if (this.sortByOptions.includes(sortBy)) {
```
djsmith85, line 110: Invert the if and create an early return to reduce if nesting

src/app/vault/ciphers.component.ts
```diff
@@ -80,6 +86,90 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+                this.sortBy = sortBy;
+                this.sortedDescending = true;
+            }
+
+            const self = this;
+
+            if (sortBy === 'lastEdited') {
+                this.ciphers.sort((a: CipherView, b: CipherView) => {
+                    if (self.sortedDescending) return new Date(b.revisionDate).valueOf() - new Date(a.revisionDate).valueOf(); // weird typescript thing: https://stackoverflow.com/a/60688789/7672369
+                    else return new Date(a.revisionDate).valueOf() - new Date(b.revisionDate).valueOf();
```
djsmith85, line 123: Please don't use single line if/return statement. Add curly braces
```suggestion
                    if (self.sortedDescending) {
                       return new Date(b.revisionDate).valueOf() - new Date(a.revisionDate).valueOf(); // weird typescript thing: https://stackoverflow.com/a/60688789/7672369
                    }
                    else {
                        return new Date(a.revisionDate).valueOf() - new Date(b.revisionDate).valueOf();
                    }
```

src/app/vault/ciphers.component.ts
```diff
@@ -80,6 +86,90 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+
+                    if (self.sortedDescending) {
+                        return sortingValue1.localeCompare(sortingValue2);
+                    } else {
+                        return sortingValue2.localeCompare(sortingValue1);
+                    }
+                });
+            } else if (sortBy === 'name') {
+                if (this.sortedDescending) this.ciphers.sort((a, b) => a.name.localeCompare(b.name));
+                else this.ciphers.sort((a, b) => b.name.localeCompare(a.name));
```
djsmith85, line 154: Same here, please use curly braces

djsmith85 (changes requested):
@fredrikburmester Thank you for your contribution, highly appreciated.

Requested some minor, mostly styling, changes. 

During testing though I saw that, when I had selected a sort by `owner`, it displays the owners the cipher items. If I then click on the column `name` it triggers the sort by name but the owners get removed. That might confuse users.

src/app/vault/bulk-actions.component.html
```diff
@@ -20,15 +20,15 @@
...
         <button class="dropdown-item" appStopClick (click)="selectAll(true)">
             <i class="fa fa-fw fa-check-square-o" aria-hidden="true"></i>
             {{'selectAll' | i18n}}
         </button>
         <button class="dropdown-item" appStopClick (click)="selectAll(false)">
             <i class="fa fa-fw fa-minus-square-o" aria-hidden="true"></i>
             {{'unselectAll' | i18n}}
-        </button>
+        </button> -->
```
djsmith85, line 31: This should be removed completely

djsmith85 (changes requested):

src/app/vault/ciphers.component.ts
fredrikburmester: I think 'My Vault' works fine since there is already a translation for this. 👍🏻

src/app/vault/ciphers.component.ts
fredrikburmester: Other functions pass the entire cipher so I tried to do the same. But I can change to only pass the necessary information if that's better?

fredrikburmester:
@djsmith85 Thank you for the feedback! Yes, the value in the sorting column is cleared when the user selects sort by name. This is because I am unsure what we want to display in this column when sorting by name. Do we keep the owner information or do we keep the last selected sorting method information (like lastEdited date), or something else?

djsmith85:
Hi @fredrikburmester I unfortunately hadn't previously looked at the Figma designs. Based on those I think the initial screen should start with the columns 'name' and 'owner'. The dropdown is actually more of a column selector(values: owner/dateCreated/lastEdited/lastUsed) which also initiates a sort after selecting an option.

From that I'd say that sorting by the name column should keep the 2nd column source on whatever the dropdown was set to.

The empty option and the 'name' option can then be removed from the dropdown options.

src/app/vault/ciphers.component.ts
```diff
@@ -80,6 +86,96 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+                this.sortBy = sortBy;
+                this.sortedDescending = true;
+            }
+
+            const self = this;
+
+            if (sortBy === 'lastEdited') {
+                this.ciphers.sort((a: CipherView, b: CipherView) => {
+                    if (self.sortedDescending) {
+                       return new Date(b.revisionDate).valueOf() - new Date(a.revisionDate).valueOf();
```
eliykat, line 123: `CipherView.RevisionDate` is already a `Date` object, do you need to construct a new `Date` object here? Could you just call `b.revisionDate.valueOf()`?

src/app/vault/ciphers.component.ts
```diff
@@ -80,6 +86,96 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+            } else {
+                this.sortBy = sortBy;
+                this.sortedDescending = true;
+            }
+
+            const self = this;
+
+            if (sortBy === 'lastEdited') {
+                this.ciphers.sort((a: CipherView, b: CipherView) => {
+                    if (self.sortedDescending) {
```
eliykat, line 122: I'd recommend moving this `if` condition out of the `sort` callback so that we evaluate it once only. Here it's being checked repeatedly as the list is being sorted. Same for the other sorting logic below.

src/app/vault/ciphers.component.ts
```diff
@@ -80,6 +86,96 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+                        return new Date(a.revisionDate).valueOf() - new Date(b.revisionDate).valueOf();
+                    }
+                });
+            } else if (sortBy === 'dateCreated') {
+                // TO-DO
+            } else if (sortBy === 'owner') {
+                const sortingArray = this.organizationNames;
+
+                this.ciphers.sort((a: CipherView, b: CipherView) => {
+                    let sortingValue1;
```
eliykat, line 134: You should be able to use ternary if conditions and `const` declarations here, e.g.
```typescript
const sortingValue1 = a.OrganizationId == null
   ? 'Personal'
   : sortingArray.get(a.organizationId);
```

This will let you be much more concise.

src/app/vault/ciphers.component.ts
```diff
@@ -80,6 +86,96 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+                    if (self.sortedDescending) {
+                       return new Date(b.revisionDate).valueOf() - new Date(a.revisionDate).valueOf();
+                    } else {
+                        return new Date(a.revisionDate).valueOf() - new Date(b.revisionDate).valueOf();
+                    }
+                });
+            } else if (sortBy === 'dateCreated') {
+                // TO-DO
+            } else if (sortBy === 'owner') {
+                const sortingArray = this.organizationNames;
```
eliykat, line 131: This seems unnecessary

src/app/vault/ciphers.component.ts
```diff
@@ -80,6 +86,96 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+                    }
+
+                    if (b.organizationId == null) {
+                        sortingValue2 = 'Personal';
+                    } else {
+                        sortingValue2 = sortingArray.get(b.organizationId);
+                    }
+
+                    if (self.sortedDescending) {
+                        return sortingValue1.localeCompare(sortingValue2);
```
eliykat, line 150: We have a helper function in `Utils.getSortFunction` which I think you should use here. You can look for examples in the code as to how it's used.

src/app/vault/ciphers.component.ts
```diff
@@ -80,6 +86,96 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+            this.sortBy = '';
+        }
+    }
+
+    reverseSort(): void {
+        this.sortCiphers(this.sortBy);
+    }
+
+    revisionDate(c: CipherView): string {
+        const revisionDate = c.revisionDate;
```
eliykat, line 172: Simplify to a ternary `if` condition

src/scss/tables.scss
```diff
@@ -72,6 +72,16 @@
         }
     }
 
+    thead {
+        tr > th {
+            cursor: pointer;
```
eliykat, line 77: We use `<th>` in other places as well, and we don't want to affect those. I suggest using a class here.

eliykat (changes requested):
Great work @fredrikburmester, thanks for the PR. I've left some detailed feedback below. I also noticed the following when testing:

* Clicking on the "Name" header does not update the Sort By header. e.g. if you sort by Owner, then click Name, it'll sort by Name but it still says Owner.
* Changing between folders/types (on the left hand side) will sort the ciphers by Name again. I think the sorting order should persist.
* If possible, the Sort By header should open a [bootstrap dropdown](https://getbootstrap.com/docs/4.0/components/dropdowns/) like in the Figma. (It looks like the menu that appears when you click the cog, not the system `<select>` dropdown.) This is lower priority I think, leave it until last. Ideally it should say "Sort By" if it's sorting by name (the default), but if you change it to another sorting option, it should display that option (e.g. "Owner") as in the Figma.

src/app/vault/ciphers.component.ts
fredrikburmester: I'm only using the set of organization names as keys for the sorting. The ciphers is the array being sorted.

src/app/vault/ciphers.component.ts
```diff
@@ -44,6 +44,11 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
     actionPromise: Promise<any>;
     userHasPremiumAccess = false;
 
+    organizationNames = new Map<string, string>();
+    sortedDescending = true;
+    sortBy: string = '';
+    readonly sortByOptions: string[] = ['name', 'owner', 'dateCreated', 'lastEdited'];
```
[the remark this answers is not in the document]
fredrikburmester: I removed this completely since it seems unnecessary to check the sorting method when the methods are hardcoded.

src/app/vault/ciphers.component.ts
fredrikburmester: I'm actually not sure how to correctly implement this so help here would be appreciated.

fredrikburmester:
Sorry for taking some time. I've now changed the behaviour to better match the improved design that was proposed by Danielle. 

Improvements include: Sorting by name does not change the data displayed in the right column, sorting is not done when selecting the data only when pressing the sorting arrows. Arrows are invisible until hovering on the table header. There is no longer a 'sort by' option, default is 'owner', but sorting is still by 'name' as default. 

Also, the code has been improved thanks to all you guys great comments @eliykat @djsmith85  

There is only a couple of things left (unresolved in the comments above @eliykat).

eliykat:
@fredrikburmester I'm still working through this one, thought I would merge with `master` for you given that there have been some major changes lately with Prettier + account switching which I'm more familiar with. Will get back to you early next week once I've finished that process and reviewed.

src/app/vault/ciphers.component.ts
```diff
@@ -81,6 +89,79 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+          b.organizationId == null ? "Personal" : this.organizationNames.get(b.organizationId);
+
+        if (this.sortedDescending) {
+          return sortingValue1.localeCompare(sortingValue2);
+        } else {
+          return sortingValue2.localeCompare(sortingValue1);
+        }
+      });
+      this.showData = "owner";
+    } else if (sortBy === "name") {
```
eliykat, line 144: I found that `CipherService` already has a `getLocaleSortingFunction` which will sort ciphers by name (with some extra logic for login items). You'll need to use that here. e.g. `this.ciphers.sort(this.cipherService.getLocaleSortingFunction)`.

To order it descending, I think you could just do the above and then `this.ciphers.reverse()`. It's simple, and [this SO post](https://stackoverflow.com/a/52030227) suggests that it's not as inefficient as you might think. But please test it and make sure it works as expected.

src/app/vault/ciphers.component.ts
```diff
@@ -81,6 +89,79 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+      if (this.sortedDescending) {
+        this.ciphers.sort((a, b) => a.name.localeCompare(b.name));
+      } else {
+        this.ciphers.sort((a, b) => b.name.localeCompare(a.name));
+      }
+    }
+  }
+
+  reverseSort(): void {
```
eliykat, line 153: This is not referenced anywhere and can be deleted.

src/app/vault/ciphers.component.ts
```diff
@@ -81,6 +89,79 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+      if (this.sortedDescending) {
+        this.ciphers.sort((a: CipherView, b: CipherView) => {
+          return b.revisionDate.valueOf() - a.revisionDate.valueOf();
+        });
+      } else {
+        this.ciphers.sort((a: CipherView, b: CipherView) => {
+          return a.revisionDate.valueOf() - b.revisionDate.valueOf();
+        });
+      }
+      this.showData = "lastEdited";
```
eliykat, line 127: I don't think you need this line (and the equivalent in each `else if` block) because we set `showData` separately now. You're only ever sorting the data that you've already displayed.

src/app/vault/ciphers.component.ts
eliykat: You might consider always sorting ascending, and then calling `reverse()` at the end if you need it descending. Would clean up this method a lot.

src/app/vault/ciphers.component.ts
eliykat: I found some cipher sorting functions in `cipherService` that I didn't know about 😬  I've included more information and an example in my new review (incoming shortly).

src/app/vault/ciphers.component.ts
eliykat: Right you are. I think your new code is much cleaner and understand here 👍

src/app/vault/ciphers.component.ts
```diff
@@ -37,6 +38,11 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
   actionPromise: Promise<any>;
   userHasPremiumAccess = false;
 
+  organizationNames = new Map<string, string>();
+  sortedDescending = true;
```
eliykat, line 42: The default sort order should be ascending, that is, A to Z. (I had to look this up because ascending/descending is confusing 😂 )

src/app/vault/ciphers.component.html
```diff
@@ -1,6 +1,35 @@
...
+            <tr>
+            <th scope="col">
+                <input (click)="selectAll()" type="checkbox">
+            </th>
+            <th scope="col"></th>
+            <th scope="col" class="user-select-none" (click)="sortCiphers('name')">Name</th>
+            <th scope="col" class="table-col-sort cipher-sort-table-header">
+                <div class="input-group m-0 font-weight-bold">
+                    <div (click)="reverseSort()" *ngIf="sortedDescending; else elseBlock" class="input-group-prepend no-bg">
+                        <label  class="input-group-text bg-transparent border-0" for="inputGroupSelect01">↓</label>
```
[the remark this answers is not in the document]
eliykat: @fredrikburmester you haven't made this change, did you have any questions or want to discuss?

src/app/vault/ciphers.component.html
```diff
@@ -7,6 +7,70 @@
...
+          </th>
+        </ng-template>
+        <th
+          scope="col"
+          colspan="2"
+          class="show-arrow-on-hover table-col-sort cipher-sort-table-header"
+        >
+          <div class="input-group m-0 font-weight-bold">
+            <div
+              (click)="sortCiphers(option.value)"
```
eliykat, line 45: Instead of referring to `option.value`, refer to the class property that control is bound to. e.g. `showData` (or whatever you rename it to)

src/app/vault/ciphers.component.html
```diff
@@ -7,6 +7,70 @@
...
+            </div>
+            <ng-template #elseBlock2>
+              <div (click)="sortCiphers(option.value)" class="input-group-prepend">
+                <label class="input-group-text bg-transparent border-0" for="sortOptions"
+                  ><i class="sorting-arrow bwi bwi-chevron-up" aria-hidden="true"></i
+                ></label>
+              </div>
+            </ng-template>
+            <select
+              #option
```
eliykat, line 61: Get rid of this id (`#option`) - see other comment. You should refer to the class property this control is bound to, not the control itself.

src/app/vault/ciphers.component.html
```diff
@@ -7,6 +7,70 @@
...
+              (change)="changeDataToShow(option.value)"
```
eliykat, line 62: You can just bind this to the class property directly rather than calling a setter method.

src/app/vault/ciphers.component.ts
```diff
@@ -81,6 +89,79 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
...
+  reverseSort(): void {
+    this.sortCiphers(this.sortBy);
+  }
+
+  changeDataToShow(type: string): void {
+    this.showData = type;
+  }
+
+  revisionDate(date: Date): string {
+    return date ? date.toLocaleDateString("en-US") : null;
```
eliykat, line 162: We should get the user's locale from `i18nService` and use that instead of assuming they want it in en-US format.

Style: we use explicit null checks (`date != null`).

src/app/vault/ciphers.component.ts
```diff
@@ -37,6 +38,11 @@ export class CiphersComponent extends BaseCiphersComponent implements OnDestroy
   actionPromise: Promise<any>;
   userHasPremiumAccess = false;
 
+  organizationNames = new Map<string, string>();
+  sortedDescending = true;
+  sortBy: string = "name";
+  showData: string = "owner";
```
eliykat, line 44: Use a more descriptive name: e.g. `selectedColumn`.

eliykat (changes requested):
I've done the following to bring you up-to-date:
1. Merged with master and taken care of all conflicts
2. Updated to use the new icon set (we now have our own `bwi` icons instead of `fa`)

I've reviewed it and it's looking good, an improvement on last time and the new design makes more sense as well. However there are still some changes required, see my further comments below.

EDIT: I messed something up in Github... my comments are both above and below 😂

Hinton:
Hi,

We are currently in the process of archiving this repository and moving the code and future development to the [bitwarden/clients](https://github.com/bitwarden/clients) repository. This PR will be closed but feel free to re-open it again on [bitwarden/clients](https://github.com/bitwarden/clients).

PR #1280, opened 2021-11-08
Add credit card logos to allow displaying icons based on brand

linusaarnio:
This is PR 4/4 related to https://community.bitwarden.com/t/credit-card-logos-for-card-vault-items/33267

## Type of change
- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective
Add images for the credit card logos.

## Code changes
This PR adds the images needed by https://github.com/bitwarden/jslib/pull/537.
They are a modified (changed size to look better in the 20x20 box for icons) version of https://www.figma.com/community/file/880472656109554171 which is licenced under [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/) by Arthur Chayka

## Screenshots

web

## Testing requirements
Test that images are shown correctly for the different brands.
Test that no brands are shown when "Disable Website Icons" is checked in options.

## Before you submit


Hinton:
@linusaarnio The icon license requires a bit of consideration, will have to check internally how to provide the required attribution or if there are alternative icons we can use instead.

linusaarnio:
@Hinton Yes it does, forgot to write about that. It is up to you how you choose to add attribution, my suggestion would be to add it in the "About" part of settings in the respective clients.

danielleflinn:
Hi @linusaarnio, thank you for your patience while we reviewed your PR. We have developed a set of icons that does not require attribution that we'd like to use for this project. See the attached zip file for compressed pngs. 

There is a light and dark version of each icon that should be used with the corresponding themes. 

The images should be sized to not stretch when used; but let me know if there are any issues when you swap out with the icons you currently have. Thanks for your contribution!

[tinified.zip](https://github.com/bitwarden/web/files/7584807/tinified.zip)

linusaarnio:
Thank you @danielleflinn, the icons look great!

Should the dark theme icons be used for all themes except for default/light one ( i.e. solarized dark and nord)?

linusaarnio:
@danielleflinn I have done the switch as I suggested in my earlier comment, with the themes Solarized dark and Nord (in the clients that have these themes) using the dark version of the icons since it seemed like the most logical option. Let me know if this should be changed.

Also, the American Express icon looks like the text is shifted a bit to the right. Is this intentional?

Hinton:
@danielleflinn It seems that discovery light is slightly bigger than the other icons. Should we update it to be consistent?

@linusaarnio I wonder slightly if we should put the icons and css in jslib to avoid duplicating them across our projects.

Hinton:
@linusaarnio we've updated the icons to have a bit better resolution. (24 x 19)

[payment-icons.zip](https://github.com/bitwarden/web/files/7686852/payment-icons.zip)

src/scss/styles.scss
```diff
@@ -1,4 +1,7 @@
 @import "../../jslib/angular/src/scss/webfonts.css";
+$dark-icon-themes: 'theme_dark';
```
Hinton, line 2: I think we could move this to variables.

src/scss/styles.scss
```diff
@@ -1,4 +1,7 @@
...
+@import "../../jslib/angular/src/scss/icons.scss";
```
Hinton, line 3: And this needs to be placed below fontawesome, since we override one of the css properties.

Hinton (changes requested):
Once jslib is merged, we'll need to update the jslib submodule.

djsmith85 (approved):
This is looking good. @linusaarnio Nice work!

Hinton (approved):

Hinton:
Thanks for the contribution @linusaarnio.

PR #1301, opened 2021-11-15
Disable key connector when org doesn't have the feature

Hinton:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Disabled the key connector option in SSO when org doesn't have the feature.

Resolves https://app.asana.com/0/1201363553288890/1201363553288934
Depends on https://github.com/bitwarden/jslib/pull/551

## Before you submit

### DIFF
--- a/bitwarden_license/src/app/organizations/manage/sso.component.html
+++ b/bitwarden_license/src/app/organizations/manage/sso.component.html
@@ -12,6 +12,8 @@
         {{'ssoPolicyHelpStart' | i18n}}
         <a routerLink="../policies">{{'ssoPolicyHelpLink' | i18n}}</a>
         {{'ssoPolicyHelpEnd' | i18n}}
+        <br>
+        {{'ssoPolicyHelpKeyConnector' | i18n}}
     </p>
     
     <div class="form-group">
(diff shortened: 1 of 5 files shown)


bitwarden_license/src/app/organizations/manage/sso.component.html
```diff
@@ -32,9 +32,14 @@ <h1>{{'singleSignOn' | i18n}}</h1>
             </label>
         </div>
         <div class="form-check mt-2 form-check-block">
-            <input class="form-check-input" type="radio" id="memberDecryptionKey" [value]="true" formControlName="useKeyConnector">
+            <input class="form-check-input" type="radio" id="memberDecryptionKey" [value]="true" formControlName="useKeyConnector"
+                [attr.disabled]="organization.useKeyConnector || null">
```
eliykat, line 36: You said you did this because you only want to disable a single radio button. But what is the point of disabling a single radio button when there are only 2 options? It seems that we could just disable the entire FormControl and the UX would be no different, and it would be a lot cleaner.

eliykat (changes requested):
See comment below. I also expect this will need to be updated with changes arising from the jslib review.

bitwarden_license/src/app/organizations/manage/sso.component.html
Hinton: We decided to leave it as is.

eliykat (commented):
Minor UI feedback: if I am not using Key Connector, neither Member Decryption option is currently selected on page load. Master Password should be selected as the default option.

Screen Shot 2021-11-17 at 10 18 42 am

Other than that, looks good to me, just needs the jslib update (which has been approved).

Hinton:
@eliykat That's due to the server side changes, it will be fixed once server is merged.

eliykat (approved):

PR #1302, opened 2021-11-16
Update jslib and use new UserVerificationService pattern

eliykat:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Follow up to https://github.com/bitwarden/jslib/pull/544, where we decided to throw errors from `UserVerificationService` and catch them in the components if required.

## Code changes

* **src/app/services/services.module.ts** - update service dependencies. Change `useClass` to `useValue` for `userVerificationService`, because we removed the `@Injectable` decorator from the base component.
* all other files: move the call to `userVerificationService` into the nearby try/catch block, and display any error that occurs to the user.

https://github.com/bitwarden/jslib/pull/544 must be merged first.

## Testing requirements

Regression testing on verify password/OTP components.

## Before you submit

### DIFF
--- a/src/app/organizations/settings/delete-organization.component.ts
+++ b/src/app/organizations/settings/delete-organization.component.ts
@@ -26,8 +26,8 @@ export class DeleteOrganizationComponent {
 
     async submit() {
         try {
-            this.formPromise = this.userVerificationService.buildRequest(this.masterPassword).then(request =>
-                this.apiService.deleteOrganization(this.organizationId, request));
+            this.formPromise = this.userVerificationService.buildRequest(this.masterPassword)
+                .then(request => this.apiService.deleteOrganization(this.organizationId, request));
             await this.formPromise;
             this.toasterService.popAsync('success', this.i18nService.t('organizationDeleted'),
                 this.i18nService.t('organizationDeletedDesc'));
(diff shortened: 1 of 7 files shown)


src/app/organizations/settings/delete-organization.component.ts
```diff
@@ -22,10 +25,16 @@ export class DeleteOrganizationComponent {
...
 
     async submit() {
-        const request = await this.userVerificationService.buildRequest(this.masterPassword);
+        let request: SecretVerificationRequest;
+        try {
+            request = await this.userVerificationService.buildRequest(this.masterPassword);
+        } catch (e) {
+            this.platformUtilsService.showToast('error', this.i18nService.t('errorOccurred'), e.message);
+            return;
+        }
```
Hinton, line 37: I wonder if we should change `formPromise` to something like:

```ts
this.formPromise = async () => {
    const request = await this.userVerificationService.buildRequest(this.masterPassword);
    return await this.apiService.deleteOrganization(this.organizationId, request);
}
```

Should let us re-use the same try catch. Thoughts?

Hinton (commented):
Build seems to be failing, had a small thought on how we could simplify things.

src/app/organizations/settings/delete-organization.component.ts
eliykat: I think this could work, I'll give it a try tomorrow

Hinton (approved):

PR #1291, opened 2021-11-11
[Key Connector] Test that Key Connector URL can be reached before saving

eliykat:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other

## Objective

Before saving Key Connector config, ping the Key Connector URL to make sure it's alive.

## Code changes

* **bitwarden_license/src/app/organizations/manage/sso.component.ts** - ping the Key Connector URL on save, return an error toast if it can't be reached.

Depends on https://github.com/bitwarden/jslib/pull/543.

## Testing requirements

Try saving the Key Connector config with valid and invalid Key Connector URLs. If the URL is invalid (i.e. Key Connector cannot be reached), you should get an error toast.

## Before you submit
- [x] I have checked for **linting** errors (`npm run lint`) (required)
- [ ] I have added **unit tests** where it makes sense to do so (encouraged but not required)
- [ ] This change requires a **documentation update** (notify the documentation team)
- [ ] This change has particular **deployment requirements** (notify the DevOps team)

PR #1308, opened 2021-11-21
[Key Connector] Fix Key Connector Url test

eliykat:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix bug from https://github.com/bitwarden/web/pull/1291: we were testing the Key Connector URL even if the Key Connector was not enabled. Also some of the logic in `submit()` was wrong.

## Code changes

* **bitwarden_license/src/app/organizations/manage/sso.component.ts**:
  * only test Key Connector URL if it's enabled
  * fix logic so that we don't test the URL again if it's already been tested and hasn't changed since then.

## Before you submit

### DIFF
--- a/bitwarden_license/src/app/organizations/manage/sso.component.ts
+++ b/bitwarden_license/src/app/organizations/manage/sso.component.ts
@@ -147,15 +147,16 @@ export class SsoComponent implements OnInit {
             return;
         }
 
-        this.keyConnectorUrl.markAsPristine();
         try {
             await this.apiService.getKeyConnectorAlive(this.keyConnectorUrl.value);
         } catch {
             this.keyConnectorIsValid = false;
+            this.keyConnectorUrl.markAsPristine();
             return;
         }
 
         this.keyConnectorIsValid = true;
+        this.keyConnectorUrl.markAsPristine();
     }
 
     get enableTestKeyConnector() {
--- a/bitwarden_license/src/app/organizations/manage/sso.component.ts
+++ b/bitwarden_license/src/app/organizations/manage/sso.component.ts
@@ -126,7 +126,7 @@ export class SsoComponent implements OnInit {
             this.enabled.setValue(response.enabled);
 
             this.platformUtilsService.showToast('success', null, this.i18nService.t('ssoSettingsSaved'));
-        } catch (e) {
+        } catch {
             // Logged by appApiAction, do nothing
         }
 
--- a/bitwarden_license/src/app/organizations/manage/sso.component.ts
+++ b/bitwarden_license/src/app/organizations/manage/sso.component.ts
@@ -126,12 +126,13 @@ export class SsoComponent implements OnInit {
         } catch (e) {
             // Logged by appApiAction, do nothing
             return;
+        } finally {
+            this.formPromise = null;
         }
 
         this.data.patchValue(response.data);
         this.enabled.setValue(response.enabled);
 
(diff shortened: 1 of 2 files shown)


bitwarden_license/src/app/organizations/manage/sso.component.ts
```diff
@@ -117,9 +117,11 @@ export class SsoComponent implements OnInit {
     }
 
     async submit() {
-        if (!this.keyConnectorIsValid || this.keyConnectorUrl.dirty)
-        {
-            await this.testKeyConnector();
+        if (this.data.get('keyConnectorEnabled').value) {
+            if (this.keyConnectorUrl.dirty || this.keyConnectorIsValid == null) {
```
Hinton, line 121: I think we need to check if `keyConnectorEnabled` is dirty as well, since someone might have a url set but disabled and then turn it on.

Hinton:
Couple of comments.

* The `testKeyConnector` needs to be called within the `formPromise`, since there's currently no user feedback that things are happening.
* Having a valid key connector url, and clicking test shows "Success", changing it and clicking test will show "Success" again until the request times out and displays the error. We shouldn't show the message until the request is done.

eliykat:
Further code changes:

* use reactive form state instead of the `keyConnectorIsValid` bool. This lets us track error and pending states without having to add new class properties, and is generally a better use of reactive forms. (Note: I have tried making this a "real" async validator but it doesn't really fit the Angular forms API so I decided against it.)
* add spinner on the Test button if validation is pending
* only mark the URL control as pristine after validation is complete, to avoid premature error messages.
* mark the URL control as dirty immediately to avoid premature error messages
* move all `apiService` calls into their own method, which is then assigned to `this.formPromise`. There was too much logic here to just chain promises in a way that would be clean and readable.

bitwarden_license/src/app/organizations/manage/sso.component.ts
```diff
@@ -117,45 +118,56 @@ export class SsoComponent implements OnInit {
...
     async submit() {
+        this.formPromise = this.postData();
+
+        let response: OrganizationSsoResponse;
+        try {
+            response = await this.formPromise;
+        } catch (e) {
+            // Logged by apiService, do nothing
+            return;
+        }
```
Hinton: If we're just going to catch and do nothing, what's the benefit vs letting the error propagate?

eliykat: Good point, I moved this into a `finally` block

Hinton (approved):

Hinton:
Would have been nice to move the logic into a separate component, but might be more effort than it's worth.

bitwarden_license/src/app/organizations/manage/sso.component.ts
```diff
@@ -117,43 +118,57 @@ export class SsoComponent implements OnInit {
...
+            // Logged by appApiAction, do nothing
+            return;
+        } finally {
+            this.formPromise = null;
+        }
+
+        this.data.patchValue(response.data);
+        this.enabled.setValue(response.enabled);
+
+        this.platformUtilsService.showToast('success', null, this.i18nService.t('ssoSettingsSaved'));
```
Hinton, line 136: We could also move this logic into the try and get rid of the catch?

Hinton (approved):

PR #1311, opened 2021-11-23
Fix basePrice to reflect the sponsorship

justindbaur:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Refactored the organization plan component so that it would reflect the actual price of the organization. Since they are getting a sponsored families plan the basePrice is dropped off and now all calculations on the screen are correct for what they will be responsible for paying.

I also ran linting that I hadn't ran it after I made PR changes.

## Screenshots

image

image

## Testing requirements

Validate that the proper price is displayed throughout the screen.

## Before you submit

### DIFF
--- a/src/app/settings/organization-plans.component.html
+++ b/src/app/settings/organization-plans.component.html
@@ -158,7 +158,7 @@
             <label class="form-check-label" for="interval{{selectablePlan.type}}">
                 <ng-container *ngIf="selectablePlan.isAnnual">
                     {{'annually' | i18n}}
-                    <small *ngIf="selectablePlan.basePrice && !acceptingSponsorship">
+                    <small *ngIf="selectablePlan.basePrice">
                         {{'basePrice' | i18n}}: {{ selectablePlan.basePrice / 12 | currency:'$'}} &times; 12
                         {{'monthAbbr' | i18n}}
                         =
--- a/src/app/settings/organization-plans.component.html
+++ b/src/app/settings/organization-plans.component.html
@@ -158,7 +158,7 @@
             <label class="form-check-label" for="interval{{selectablePlan.type}}">
                 <ng-container *ngIf="selectablePlan.isAnnual">
                     {{'annually' | i18n}}
-                    <small *ngIf="selectablePlan.basePrice">
+                    <small *ngIf="selectablePlan.basePrice && !acceptingSponsorship">
                         {{'basePrice' | i18n}}: {{ selectablePlan.basePrice / 12 | currency:'$'}} &times; 12
                         {{'monthAbbr' | i18n}}
                         =
--- a/src/app/settings/organization-plans.component.html
+++ b/src/app/settings/organization-plans.component.html
@@ -166,10 +166,10 @@
                         /{{'year' | i18n}}
                     </small>
                     <small *ngIf="acceptingSponsorship">
-                        {{'basePrice' | i18n}}: {{ 40 / 12 | currency:'$'}} &times; 12
+                        {{'basePrice' | i18n}}: {{ selectablePlan.basePrice / 12 | currency:'$'}} &times; 12
                         {{'monthAbbr' | i18n}}
                         =
-                        <span style="text-decoration: line-through;">{{40 | currency:'$'}}</span>
+                        <span style="text-decoration: line-through;">{{selectablePlan.basePrice | currency:'$'}}</span>
                         {{'freeWithSponsorship' | i18n}}
                     </small>
                     <small *ngIf="selectablePlan.hasAdditionalSeatsOption">
--- a/src/app/settings/organization-plans.component.html
+++ b/src/app/settings/organization-plans.component.html
@@ -158,12 +158,19 @@
             <label class="form-check-label" for="interval{{selectablePlan.type}}">
                 <ng-container *ngIf="selectablePlan.isAnnual">
                     {{'annually' | i18n}}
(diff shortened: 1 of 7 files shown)


MGibson1 (commented):
I'm not sure this is the right way to do this. It implies the user will not be responsible for the cost should the sponsorship be revoked.

I think we either keep the prices as-is and display a note that they won't be charged for the plan as long as they're sponsored. Or we do this and add a note that the actual cost is $40/yr and they'll be charged if they lose their sponsorship

justindbaur:
@MGibson1 Would you propose the text under `Payment Information` be changed to include "... or your sponsorship expires  in which case the price will be an additional $40."?

patrick-bitwarden:
Can we get Product to weigh in on the copy question?

danielleflinn:
This looks good! Only adjustment, instead of hiding the "Base Price" line in the Summary, let's add a strikethrough to the "$40" and "FREE with Sponsorship" following the price. See [Figma for mock.](https://www.figma.com/file/x9bNsJmSss2JCpqVXY4fbV/Free-Bitwarden-Families?node-id=271%3A15209)

src/app/settings/organization-plans.component.html
```diff
@@ -165,6 +165,13 @@ <h2 class="spaced-header">{{'summary' | i18n}}</h2>
                         {{selectablePlan.basePrice | currency:'$'}}
                         /{{'year' | i18n}}
                     </small>
+                    <small *ngIf="acceptingSponsorship">
+                        {{'basePrice' | i18n}}: {{ 40 / 12 | currency:'$'}} $times; 12
```
MGibson1, line 169: This $40 should be the `basePrice` so that we can update the price as we want

src/app/settings/organization-plans.component.html
```diff
@@ -165,6 +165,13 @@ <h2 class="spaced-header">{{'summary' | i18n}}</h2>
                         {{selectablePlan.basePrice | currency:'$'}}
                         /{{'year' | i18n}}
                     </small>
+                    <small *ngIf="acceptingSponsorship">
```
MGibson1, line 168: It looks like you're missing this bool flag on the above `ngIf` as well

src/app/settings/organization-plans.component.html
MGibson1: The screenshot shows `&times;` rather than `x`, not sure why.

MGibson1 (changes requested):

src/app/settings/organization-plans.component.html
justindbaur: I did have it there but it's not needed as the basePrice for the sponsorship is 0. Therefore it's already `falsey` and doesn't display. If you want it there for verbosity though and I can add it.

src/app/settings/organization-plans.component.html
justindbaur: I change `basePrice` to 0 so that the rest of the calculations will show the correct number. I don't love hardcoding it either though. But to not hardcode it we would need some sort of `originalBasePrice` on the `PlanResponse`.

src/app/settings/organization-plans.component.html
justindbaur: Added it above as well since we are no longer changing the `basePrice` to 0 and we need it isn't `falsey`.

src/app/settings/organization-plans.component.html
justindbaur: No longer doing this.

src/app/settings/organization-plans.component.html
```diff
@@ -158,13 +158,20 @@ <h2 class="spaced-header">{{'summary' | i18n}}</h2>
...
+                        {{'basePrice' | i18n}}: {{ selectablePlan.basePrice / 12 | currency:'$'}} &times; 12
+                        {{'monthAbbr' | i18n}}
+                        =
+                        <span style="text-decoration: line-through;">{{selectablePlan.basePrice | currency:'$'}}</span>
+                        {{'freeWithSponsorship' | i18n}}
+                    </small>
```
MGibson1, line 174: Can't we just have one `<small>` node and alter the right hand side of the equals depending on the `acceptingSponsorship` boolean?

justindbaur: Switched to that.

MGibson1 (approved):
Very nice job, this ended up being way nicer than I feared

src/app/settings/organization-plans.component.html
```diff
@@ -92,7 +92,7 @@ <h2 class="mt-5">{{'chooseYourPlan' | i18n}}</h2>
                 </small>
             </ng-template>
             <span *ngIf="selectableProduct.product != productTypes.Free">
-                <ng-container *ngIf="selectableProduct.basePrice">
+                <ng-container *ngIf="selectableProduct.basePrice && !acceptingSponsorship">
```
MGibson1, line 95: I don't think this should have `acceptingSponsorship`, right?

MGibson1 (changes requested):

MGibson1 (approved):

src/app/settings/organization-plans.component.html
justindbaur: This one **does** need it. Otherwise it will show " $3.33 /month, includes 6 users" which wouldn't be true.

PR #1313, opened 2021-11-24
BEEEP: Refactor services DI

Hinton:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

We currently initialize all our services manually in `ServicesModule`. This makes it a bit cumbersome since we have to define everything in order, and should make it easier to detect bad patterns like circular dependencies.

~~This should also open up for us to define the core dependencies in `jslib` and only define what we need to override in each application.~~ I've moved most of the DI initialization to jslib, https://github.com/bitwarden/jslib/pull/565.

## Testing requirements

We should be a bit vigilant during regression testing and verify everything works.

## Before you submit

### DIFF
--- a/src/app/accounts/lock.component.ts
+++ b/src/app/accounts/lock.component.ts
@@ -1,4 +1,7 @@
-import { Component } from '@angular/core';
+import {
+    Component,
+    NgZone,
+} from '@angular/core';
 import { Router } from '@angular/router';
 
 import { ApiService } from 'jslib-common/abstractions/api.service';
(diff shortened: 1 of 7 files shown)


eliykat (commented):
This all looks good to me, it makes total sense that jslib should export a fully configured module rather than individual services. I also like that we let Angular create all the dependencies instead of instantiating everything by hand. Overall this should reduce boilerplate and maintenance costs between jslib and the clients.

Making circular dependencies more visible is a feature, not a bug. We should feel bad that we have them. :)

I think there's a service from jslib that you should move here (see my comment on the jslib PR), then I'm happy to approve. It might be a good idea to implement it for the other clients as well, before we merge, just to make sure there are no gotchas between the different environments.

eliykat (approved):

PR #1372, opened 2022-01-06
WIP: Tailwind experiment

Hinton:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

This is an example of how tailwind can be configured to run side-by-side with our existing styling. It's mostly for demonstration purposes.

## Before you submit


bitwarden_license/src/app/main.ts
```diff
@@ -7,6 +7,7 @@ import "popper.js";
 
 // tslint:disable-next-line
 require("src/scss/styles.scss");
+require("src/scss/tailwind.css");
```
Hinton, line 10: Adding tailwind as a separate `css` file prevents sass from having to re-compile all the time. And cuts down the compile time from 20s to 1s.

src/scss/tailwind.css
```diff
@@ -0,0 +1,43 @@
...
+  --color-text-body: 33 37 41;
+  --color-text-muted: 108 117 125;
+  --color-text-black50: 0 0 0 0.5;
+  --color-text-white50: 255 255 255 0.5;
+
+  --color-primary-btn-hover: #134eb9;
+  --color-primary-btn-text: #fff;
+
+  --color-secondary-btn-hover: #b8c1ca;
+  --color-secondary-btn-text: #212529;
```
Hinton, line 21: Tailwind doesn't natively support themes but rather uses variants, primarily `dark:` to specify how a component should look in dark mode. As we want to be able to support additional themes beyond light and dark, we need to extend the theming which is easiest done by using css variables.

tailwind.config.js
```diff
@@ -0,0 +1,45 @@
...
+      pattern: /tw-text-(primary|secondary)-btn-text/,
+    },
+    {
+      pattern: /tw-bg-(primary|secondary)/,
+      variants: ["hover"],
+    },
+  ],
+  theme: {
+    colors: {
+      primary: withOpacityValue("--color-primary"),
```
Hinton, line 27: I would prefer investigating if we could go towards using a color palette instead of hard-coding single-use variables like `primary-btn-hover`. Ideally we would be able to use `primary-500` for backgrounds, `primary-600` for borders, and maybe `primary-700` for hover state. Which should allow us to minimize the unique variables.

tailwind.config.js
```diff
@@ -0,0 +1,45 @@
...
+      pattern: /tw-(border|bg)-(primary|secondary)/,
+    },
+    {
+      pattern: /tw-text-(primary|secondary)-btn-text/,
+    },
+    {
+      pattern: /tw-bg-(primary|secondary)/,
+      variants: ["hover"],
+    },
+  ],
```
Hinton, line 24: Some components like `button` might want to dynamically generate classes. Since tailwind uses a JIT these won't be picked up automatically. The alternative is to explicitly define them which might be preferred. But I wanted to highlight that we can safelist items using regex as well.

tailwind.config.js
dltmurphy: Unique variables were requested so that it was easier to maintain and fix bugs after the initial release of dark mode. Now that most issues are fixed it should be quite simple to switch to this method for them instead

PR #1384, opened 2022-01-10
Add Inputmode for tel and email (identities)

coliff:
helps improve the UX for users with virtual keyboards (mobile devices and tablets

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Helps improve the UX for users with virtual keyboards (mobile devices and tablets).

## Code changes

This PR supercedes: bitwarden#661

I wrote about inputmode for CSS Tricks:
https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/#aa-tel
https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/#aa-email

## Before you submit

### DIFF
--- a/src/app/vault/add-edit.component.html
+++ b/src/app/vault/add-edit.component.html
@@ -587,6 +587,7 @@
                 id="idEmail"
                 class="form-control"
                 type="text"
+                inputmode="email"
                 name="Identity.Email"
                 [(ngModel)]="cipher.identity.email"
                 appInputVerbatim
@@ -599,6 +600,7 @@
                 id="idPhone"
                 class="form-control"
                 type="text"
+                inputmode="tel"
                 name="Identity.Phone"
                 [(ngModel)]="cipher.identity.phone"
                 [disabled]="cipher.isDeleted || viewOnly"


Hinton:
@danielleflinn Any thoughts from the design aspects? My primary concern is that it won't be possible to input non-numeric characters in the phone input. But I'm not sure why anyone would do that in the first place.

I took a look at our mobile app, and inputmode for email seems to be email, but text for phone.

Hinton (approved):

danielleflinn:
It is a standard UX pattern to restrict input for a phone field to just numbers. The only potential concern I can think of is I don't believe we are formatting that field currently so any entry would be difficult to read since it'd show up as "5554443333".

Ideally we'd format the input number automatically to "555-444-3333"; however being an international product, this introduces it's own challenges in identifying what phone number format to use by region.

All in all, I do agree that only allowing numerical digits in the phone fields does improve the experience even without the ability to format the numbers.

djsmith85:
We could use https://github.com/google/libphonenumber for parsing and formatting. There's also a port of it in C# which could be used on mobile.

Hinton:
@djsmith85 Difficult thing is figuring out what region a number is for. (I guess we have a country field in identity that could be used for displaying though)

Hinton:
Merged, thanks for the contribution @coliff.

PR #1412, opened 2022-01-27
Fix launch of webvault with latest jslib

djsmith85:
## Type of change

- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
With a change made in environment.service in jslib we had issues starting up the webvault with latest jslib in dev mode.

## Code changes

- **src/app/services/services.module.ts:** Pass true instead of false to the environmentService.setUrls to persist the settings (urls)

## Testing requirements
The web vault should start up with any environment set (dev, qa, prod)

## Before you submit

### DIFF
--- a/src/app/services/services.module.ts
+++ b/src/app/services/services.module.ts
@@ -75,7 +75,7 @@ export function initFactory(
 
     const urls = process.env.URLS as Urls;
     urls.base ??= window.location.origin;
-    environmentService.setUrls(urls, false);
+    environmentService.setUrls(urls, true);
 
     setTimeout(() => notificationsService.init(), 3000);
 


src/app/services/services.module.ts
```diff
@@ -75,7 +75,7 @@ export function initFactory(
 
     const urls = process.env.URLS as Urls;
     urls.base ??= window.location.origin;
-    environmentService.setUrls(urls, false);
+    environmentService.setUrls(urls, true);
```
Hinton, line 78: I still don't like that we're suddenly saving urls for no "reason".

Hinton: If all clients defaults to true, let's just remove it.

PR #1418, opened 2022-01-31
Fix various bugs in Options page

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

1. When saving on the options page, the confirmation toast has duplicate text in it. It says 'Options updated.' twice within the same toast.
2. In Firefox, the locale dropdown would not have a default setting. In Chrome, the page would always refresh (and then lock the vault) when saving.

## Code changes

- **src/app/settings/options.component.ts** 
  -  remove duplicate text
  - set the `startingLocale` and `startingTheme` properties properly. These should be set to the same value as `locale` and `theme`, but [this change](https://github.com/bitwarden/web/commit/17ae5ee57c4ecd1ae59b9519b5727bb2546061cb#diff-4c73005a246b799eef390729c9ec4364f3e45723c1bc351a511235caae689ee3R80) replaced the multiple left-hand assignment with null coalescing, which changed the behaviour. However, I think this illustrates that multiple left-hand assignment can be confusing and is best avoided.

## Before you submit

### DIFF
--- a/src/app/settings/options.component.ts
+++ b/src/app/settings/options.component.ts
@@ -74,8 +74,12 @@ export class OptionsComponent implements OnInit {
     this.disableIcons = await this.stateService.getDisableFavicon();
     this.enableGravatars = await this.stateService.getEnableGravitars();
     this.enableFullWidth = await this.stateService.getEnableFullWidth();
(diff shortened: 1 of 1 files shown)


differsthecat (approved):
Looks good 👍

PR #1426, opened 2022-02-01
Extract Connectors

Hinton:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

As a preparation for potentially moving to Angular CLI, and for cleaning up the main webpack config, this PR moves the connector logic to a separate webpack config.

During this move, I also decoupled the connectors from the regular css, and bootstrap, which drastically decreased the css files (and resolved the sass limitation of paths originating from the source file, not relative to the imported files.

Before this can be merged we need to figure out how we want to deploy and handle the connectors.

## Code changes

- **file.ext:** Description of what was changed and why

## Testing requirements

Regression test the connectors to ensure they still look good, and behave as expected.

## Before you submit

- [x] This change has particular **deployment requirements** (notify the DevOps team)


Hinton:
@joseph-flinn I think we settled on having these be copied into the current build? They would be placed in `/connectors/` instead of their corrent location at the root.

To ensure backwards compatability we can update our cloud proxying rules to support the old connector urls, as well as configure the self-hosted nginx instance.

MGibson1:
Added hold label to hold off on worrying about this until Feb release is out

joseph-flinn:
@Hinton I'll be available next week after the release work is done to assist with testing or any support work.

PR #1449, opened 2022-02-07
Send in error message for toast.

justindbaur:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

When attempting to enable a policy that requires you to have a prerequisite policy. This is from when we switched from directly using the toaster library (#1322) and now go through platform utils. The toaster library support an error object but platform utils does not. 

Asana Task: https://app.asana.com/0/1201259213711784/1201780427940282/f

## Code changes

- **src/app/organizations/manage/policy-edit.component.ts:** Platform utils expects a string/string[] so I send it the message property from an error so that it's get's a descriptive message of what went wrong. 

## Testing requirements

Ensure that you get a toast error when attempting to enable a policy that can't be enabled.

## Before you submit

### DIFF
--- a/src/app/organizations/manage/policy-edit.component.ts
+++ b/src/app/organizations/manage/policy-edit.component.ts
@@ -83,7 +83,7 @@ export class PolicyEditComponent {
     try {
       request = await this.policyComponent.buildRequest(this.policiesEnabledMap);
     } catch (e) {
-      this.platformUtilsService.showToast("error", null, e);
+      this.platformUtilsService.showToast("error", null, e.message);
       return;
     }
 


MGibson1 (approved):
I wonder how many other places we do this, though...

I wonder if it's better to just add a type `{message: string}` to our accepted test union type on PlatformUtilService.

justindbaur:
@MGibson1 Not actually in as many places as you might think. Most of the time it doesn't send the error message and sends an internationalized message. But this catch could catch any number of different errors that have their own internationalize message.

PR #1458, opened 2022-02-10
Exclude jslib from prettier hook

eliykat:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [x] Other

## Objective

Workaround for the following error caused by our precommit hooks (reported in https://github.com/okonet/lint-staged/issues/929). This occurs when updating jslib without making any other changes:

```terminal
git commit -m 'Update jslib'
✖ Preparing...
✔ Hiding unstaged changes to partially staged files...
↓ Skipped because of previous git error. [SKIPPED]
↓
  ✖ lint-staged failed due to a git error. [SKIPPED]
↓
  ✖ lint-staged failed due to a git error. [SKIPPED]
↓
  ✖ lint-staged failed due to a git error. [SKIPPED]

  ✖ lint-staged failed due to a git error.
  Any lost modifications can be restored from a git stash:

    > git stash list
    stash@{0}: automatic lint-staged backup
    > git stash apply --index stash@{0}

husky - pre-commit hook exited with code 1 (error)
```

## Code changes

Exclude jslib from the prettier precommit hook.

Once merged, I'll submit identical PRs for the other TS clients.

Reference: https://github.com/okonet/lint-staged#filtering-files
## Before you submit

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "npm": "~8"
   },
   "lint-staged": {
-    "*": "prettier --ignore-unknown --write",
+    "!(./jslib/**)*": "prettier --ignore-unknown --write",
     "*.png": "node scripts/optimize.js"
   }
 }
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "npm": "~8"
   },
   "lint-staged": {
-    "!(./jslib/**)*": "prettier --ignore-unknown --write",
+    "!(./jslib)*": "prettier --ignore-unknown --write",
     "*.png": "node scripts/optimize.js"
   }
 }
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "npm": "~8"
   },
   "lint-staged": {
-    "!(./jslib)*": "prettier --ignore-unknown --write",
+    "./!(jslib)**": "prettier --ignore-unknown --write",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -100,7 +100,7 @@
     "npm": "~8"
   },
   "lint-staged": {
-    "*": "prettier --ignore-unknown --write",
+    "!(./jslib/**)*": "prettier --ignore-unknown --write",
```
Hinton, line 103: Do we need to specify `/**` here?

Hinton (approved):
Nice find if it solves the pesky submodule errors

package.json
eliykat: Apparently not, I removed it and it still tested fine.

eliykat:
Requires further testing 🤔  the problem is that if you have any other unstaged changes, the error won't throw. So you have to commit the change to `package.json`, then test it, not the other way around. This doesn't actually solve the problem at the moment.

eliykat:
OK, here we go:
* `./` starts from git root
* `!(jslib) excludes the jslib folder
* `**` matches all other contents recursively

Confirmed that:
* fixes the bug
* still matches and runs on files in git root
* still matches and runs on files in subdirs

Hinton (approved):

PR #1470, opened 2022-02-11
[bug] Store vault data in memory

addisonbeck:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Resolve https://app.asana.com/0/1201804463288267/1201794888247580

## Code changes
We were storing vault data in memory in the web vault to avoid key size limits for storing data in Chrome. This was broken in with the storage restructure, and is put back in this PR. Instead of saving vault data to disk we save it to in memory state.

I tested this by connecting my local web client to the qa cloud vault, and used @Larry-Sussman's test account from the ticket. Vault data loads in as expected.

## Screenshots
Screen Shot 2022-02-11 at 3 14 14 AM

## Before you submit

### DIFF
--- a/src/services/state.service.ts
+++ b/src/services/state.service.ts
@@ -7,6 +7,11 @@ import { StateService as StateServiceAbstraction } from "../abstractions/state.s
 
 import { StorageOptions } from "jslib-common/models/domain/storageOptions";
 
+import { CipherData } from "jslib-common/models/data/cipherData";
+import { CollectionData } from "jslib-common/models/data/collectionData";
+import { FolderData } from "jslib-common/models/data/folderData";
+import { SendData } from "jslib-common/models/data/sendData";
+
 export class StateService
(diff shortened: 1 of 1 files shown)


src/services/state.service.ts
```diff
@@ -33,4 +38,58 @@ export class StateService
       this.reconcileOptions(options, await this.defaultOnDiskLocalOptions())
     );
   }
+
+  async getEncryptedCiphers(options?: StorageOptions): Promise<{ [id: string]: CipherData }> {
```
Hinton, line 42: I wonder if we should add the `override` key here to make it obvious we're overriding stuff. But to be fair I don't believe we do it in other places. (We should turn on the --no-implicit-override flag in TS)

Hinton (approved):
This looks good from a pure code perspective. Have we checked what happens if we do a refresh of the browser tab and attempt to unlock it again?

PR #1524, opened 2022-03-08
[End User Vault Refresh] Organizations - updated nav and route permissions

eliykat:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Implement SG-33 in the [Organization Page story](https://bitwarden.atlassian.net/browse/SG-18).
* add top-level Organization page
* update route permissions and templating logic so that ordinary users do not have access to this section anymore.

## Code changes

- **src/app/layouts/navbar.component** - add new top-level Organization button. Only show it to users who can access some part of the admin view (i.e. admin, owner, or custom permissions).

- **src/app/oss-routing.module.ts** - support navigating to `/organizations` without an orgId. In practice this will be handled by the `OrganizationGuardService` changes below.

- **src/app/services/organization-guard.service.ts** - instead of rejecting navigation to `/organizations` without a supplied Id, redirect the user to the the first org they have permission to access (in alphabetical order). This lets us keep the navbar simple and navigate to a generic route from any component without having to figure this out.

EDIT: (snip), I have removed org picker stuff which has been moved to a separate branch. See discussion below.

Depends on https://github.com/bitwarden/jslib/pull/727

## Testing requirements

General testing and regression on navigating to Organizations and around Organization tabs with various permissions. (owner, admin, user, custom permissions etc)

## Before you submit


src/app/components/organization-picker.component.html
```diff
@@ -0,0 +1,59 @@
...
+    </button>
+    <div class="dropdown-menu" aria-labelledby="pickerButton">
+      <a
+        *ngFor="let org of organizations"
+        class="dropdown-item"
+        [routerLink]="['/organizations', org.id]"
+      >
+        <i
+          class="bwi bwi-check mr-2"
+          [ngClass]="org.id === activeOrganizationId ? 'visible' : 'invisible'"
```
Hinton, line 28: Any reason we use visible here vs ngIf?

src/app/components/organization-picker.component.ts
```diff
@@ -0,0 +1,44 @@
...
+  get activeOrganization() {
+    return this.organizations.find((org) => org.id == this.activeOrganizationId);
+  }
+
+  async ngOnInit() {
+    this.route.params.subscribe(async (params: any) => {
+      this.activeOrganizationId = params.organizationId;
+      await this.load();
+    });
+  }
```
Hinton, line 34: We should avoid using routes unless there are no other option. This is not a "page" component and we can simply use `@Input()` for this which is more testable and easier to work with.

src/app/components/organization-picker.component.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  async load() {
+    const orgs = await this.organizationService.getAll();
+    this.organizations = orgs
+      .filter((org) => org.canAccessAdminView)
+      .sort(Utils.getSortFunction(this.i18nService, "name"));
+
+    this.loaded = true;
+  }
```
Hinton, line 43: If we change to using inputs, this could be `ngOnChanges`.

Hinton (changes requested):

src/app/components/organization-picker.component.html
eliykat: `invisible` hides the icon but it still takes up room in the DOM, keeping everything nicely aligned with the actual icon size.

src/app/services/organization-guard.service.ts
```diff
@@ -15,21 +16,45 @@ export class OrganizationGuardService implements CanActivate {
...
     }
+
+    if (!org.canAccessAdminView) {
+      this.platformUtilsService.showToast(
+        "error",
+        this.i18nService.t("accessDenied"),
+        this.i18nService.t("accessDeniedOrganizations")
+      );
+      return this.cancelNavigation();
+    }
```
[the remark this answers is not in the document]
eliykat: Why? We're changing who can access a certain page, if someone is trying to navigate to a page that they've suddenly lost access to, I think we should tell them why. (or get support tickets about it)

src/app/services/organization-guard.service.ts
```diff
@@ -15,21 +16,45 @@ export class OrganizationGuardService implements CanActivate {
...
+    if (route.params.organizationId == null) {
+      const allOrgs = await this.organizationService.getAll();
+      const adminOrgs = allOrgs.filter((org) => org.isAdmin);
+      if (adminOrgs.length < 1) {
+        return this.cancelNavigation();
+      }
+      const sortedOrgs = adminOrgs.sort(Utils.getSortFunction(this.i18nService, "name"));
+      this.router.navigate(["/organizations", sortedOrgs[0].id]);
+      return false;
+    }
```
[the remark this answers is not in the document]
eliykat: I've moved this into the layout component as you suggest, let me know what you think.

src/app/components/organization-picker.component.ts
eliykat: Done

src/app/components/organization-picker.component.ts
eliykat: The input is only used for the active org, and that flows through the input binding automatically.
This logic only needs to be run when the component is first loaded.

eliykat:
@Hinton I've changed how we handle access/view permissions for the Organization admin page and routes.

The problem is that we need to check whether a user can access the top-level Organization page for a given org, and we need to do this in multiple locations:
* navbar, so we know whether to display the Organizations link
* route guard, so we know whether to allow navigation
* org picker, so we know which orgs to display

In turn, whether they can access the Organization page depends on whether they can access any of the top level tabs in that page. So it's all related and needs to be kept in one place. (If we don't DRY then it's too easy for permissions to be updated in one place and not the other.)

I've centralised all of this by exporting the routing permissions from `oss-routing.module.ts`. Each component imports this and then uses a new method on the `organization` model to see whether those permissions are met. See https://github.com/bitwarden/jslib/pull/727.

There turned out to be quite a bit going on in this PR, let me know if you want me to split out the picker from the routing/nav changes.

My only remaining task is to replace the Bootstrap dropdown with Angular CDK.

src/app/services/organization-guard.service.ts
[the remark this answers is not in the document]
eliykat: Having a route-guard error handling service would be nice but I agree not a priority right now.

eliykat:
Org picker has been moved to a separate PR/branch so that we can just focus on the routing permissions here.

src/app/components/organization-picker.component.ts
eliykat: Marking this as resolved because it'll be in a separate PR

src/app/components/organization-picker.component.html
eliykat: Marking this as resolved because it'll be in a separate PR

src/app/services/organization-guard.service.ts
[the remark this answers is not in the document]
eliykat: Update: the tab pages get the `organization.id` from the route, so even if we get to the layout component, we still have to navigate to a new route to update the tab pages. I think redirection logic should go in a route guard. But I've at least split it out into a separate guard and refactored the organization route guards to reduce duplication.

eliykat:
Closed this because it was getting messy. Opened fresh PR to cover routing changes: https://github.com/bitwarden/web/pull/1551

PR #1533, opened 2022-03-10
[End User Vault Refresh] Vault - update cog menu

eliykat:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

* SG-31: Change the cog menu to 3-dots, make it always visible (not just on hover).
* While I'm here: change these menu items to use `<button>` elements instead of `<a>`. These are buttons that perform actions, not links that trigger navigation. This matters particularly for a11y.

## Code changes

Table row option menus and buttons are used throughout the app, and they're all the same. Therefore I've created the following components:
* **src/app/components/table-row-option-menu.component** - the bootstrap dropdown menu. Use it in a table row. Wrap it around the buttons you want in the menu.
* **src/app/components/table-row-option.component** - the individual buttons in the menu.

While these are small HTML snippets, they occur > 60 times throughout the app, and are a pain to update individually. They seemed like a good candidate for the component library.

All other changes implement these new components through every table in the app. 

This goes beyond just the End User Vault Refresh, however this change is also in the Org Admin Refresh work, so it makes sense to change all tables at once to keep a consistent look throughout the app.

## Screenshots

Screen Shot 2022-03-10 at 2 25 14 pm

## Testing requirements

Test all table option menus throughout the app - i.e.
* ensure no change in appearance
* ensure no change in functionality

## Before you submit


src/app/components/table-row-option-menu.component.html
```diff
@@ -0,0 +1,19 @@
+<div class="dropdown">
+  <button
+    class="btn dropdown-toggle disable-caret"
+    type="button"
+    id="dropdownMenuButton-{{ rowIndex }}"
```
Hinton, line 5: This is just a global id to track bootstrap right? I wonder if we should just take the opportunity to replace the dropdown logic with https://material.angular.io/cdk/overlay/overview

src/app/components/table-row-option-menu.component.ts
```diff
@@ -0,0 +1,9 @@
+import { Component, Input } from "@angular/core";
+
+@Component({
+  selector: "app-table-row-option-menu",
```
Hinton, line 4: I feel like we could rename this component to `app-menu`. To my knowledge there's nothing that directly couples this to a table row other than the icon? Which we could make customizable similar to https://material.angular.io/components/menu/overview

src/app/components/table-row-option.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component, Input, Output, EventEmitter } from "@angular/core";
+
+@Component({
+  selector: "app-table-row-option",
+  templateUrl: "table-row-option.component.html",
+})
+export class TableRowOptionComponent {
```
Hinton, line 7: What do you think of following the Angular Material component design with a directive here? Would allow for using links should they be needed.

eliykat:
Will open a fresh PR using `bit-menu`

PR #1536, opened 2022-03-10
Update reports page

Hinton:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Update the reports page according to spec.

Resolves SG-12

## Code changes

- **src/app/components/premium-badge.component.ts:** Add tabindex, add condition support.
- **src/app/layouts/navbar.component.html**: Add reports page to navbar
- **src/app/oss-routing.module.ts**: Move reports routing to separate routing module.
- **src/app/oss.module.ts**: Add ButtonModule to OssModule.
- **src/app/reports/report-card.component.ts**: Report card component.
- **src/app/reports/report-list.component.ts**: Add component for reports homepage.
- **src/app/reports/reports.component.ts**: Add component for report layout.
- **src/app/reports/reports.module.ts**: Add reports module for routing.
- **src/app/tools/tools.component.html**: Remove report links.

## Screenshots

image

image

image

image

Screenshot 2022-03-18 at 10 53 25

## Before you submit


src/app/reports/reports.component.ts
```diff
@@ -0,0 +1,25 @@
...
+@Component({
+  selector: "app-reports",
+  templateUrl: "reports.component.html",
+})
+export class ReportsComponent implements OnDestroy {
+  homepage = true;
+  subscription: Subscription;
+
+  constructor(router: Router) {
+    this.subscription = router.events
```
vincentsalucci, line 15: I was just researching this - seems pretty wild that this is the most definitive way to ask "where am I?" instead of what seems like more obvious ways like checking the `ActivatedRoute` or even the `Router` - neither of which have consistent ways of accessing that specific data :)

vincentsalucci, line 15: I was just researching this - seems pretty wild that this is the most definitive way to ask "where am I?" instead of what seems like more obvious ways like checking the `ActivatedRoute` or even the `Router` - neither of which have consistent ways of accessing that specific data :)

src/app/reports/reports.component.html
```diff
@@ -0,0 +1,12 @@
+<div class="container page-content">
```
vincentsalucci, line 1: Might be nice to add a screenshot of a selected report to see this button.

vincentsalucci, line 1: Might be nice to add a screenshot of a selected report to see this button.

src/app/reports/report-card.component.ts
```diff
@@ -0,0 +1,166 @@
...
+    private sanitizer: DomSanitizer
+  ) {}
+
+  async ngOnInit() {
+    this.report = reports[this.type];
+
+    this.hasPremium = await this.stateService.getCanAccessPremium();
+  }
+
+  get premium() {
```
vincentsalucci, line 145: Was a big confused at this name - but I honestly am not sure of a better one without making it very long and potentially even more confusing.

vincentsalucci, line 145: Was a big confused at this name - but I honestly am not sure of a better one without making it very long and potentially even more confusing.

src/app/reports/report-card.component.ts
```diff
@@ -0,0 +1,166 @@
...
+  constructor(
+    private stateService: StateService,
+    private messagingService: MessagingService,
+    private sanitizer: DomSanitizer
+  ) {}
+
+  async ngOnInit() {
+    this.report = reports[this.type];
+
+    this.hasPremium = await this.stateService.getCanAccessPremium();
```
vincentsalucci, line 142: Is there anyway to move this out of the card itself? Seems like calling this for every card isn't really necessary. Maybe having it in the `list` component or even the parent `reports component` and passing it down could save a feel calls to the `stateService`.

vincentsalucci, line 142: Is there anyway to move this out of the card itself? Seems like calling this for every card isn't really necessary. Maybe having it in the `list` component or even the parent `reports component` and passing it down could save a feel calls to the `stateService`.

src/app/reports/report-card.component.ts
```diff
@@ -0,0 +1,166 @@
+import { Component, Input, OnInit } from "@angular/core";
+import { DomSanitizer } from "@angular/platform-browser";
+
+import { MessagingService } from "jslib-common/abstractions/messaging.service";
+import { StateService } from "jslib-common/abstractions/state.service";
+
+export enum ReportTypes {
```
vincentsalucci, line 7: I'm confused by this enum - can you help clarify its purpose?

vincentsalucci, line 7: I'm confused by this enum - can you help clarify its purpose?

src/app/reports/report-card.component.ts
```diff
@@ -0,0 +1,166 @@
...
+  icon: string;
+  requiresPremium: boolean;
+};
+
+const reports: Record<ReportTypes, ReportEntry> = {
+  exposedPasswords: {
+    title: "exposedPasswordsReport",
+    description: "exposedPasswordsReportDesc",
+    route: "exposed-passwords-report",
+    icon: `
```
vincentsalucci, line 29: Is there any specific reason why we're storing the `svg` data inline instead of saving the image in the project and pathing to it when needed? Would doing something like that change your card architecture too much?

vincentsalucci, line 29: Is there any specific reason why we're storing the `svg` data inline instead of saving the image in the project and pathing to it when needed? Would doing something like that change your card architecture too much?

src/app/components/premium-badge.component.ts
```diff
@@ -1,17 +1,33 @@
...
+      {{ "premium" | i18n }}
+    </button>
   `,
 })
-export class PremiumBadgeComponent {
-  constructor(private messagingService: MessagingService) {}
+export class PremiumBadgeComponent implements OnInit {
+  // Optional condition defaults to premium
+  @Input() condition: boolean;
+  @Input() tabindex?: number = null;
```
vincentsalucci, line 24: What's this variables purpose? Looks like we're always setting it to `-1` when using this the `report cards`

src/app/reports/reports.component.ts
Hinton: We have to continue to listen for events since this component hangs around for each report.

src/app/reports/report-card.component.ts
Hinton: Primarily to be able to use currentColor. Only alternative is to use embeed which messes up centering of the svgs.

src/app/reports/reports.component.html
Hinton: Added.

src/app/reports/report-card.component.ts
Hinton: It's used in the `ReportListComponent`. And as a compiler check for the report type.

src/app/reports/report-card.component.ts
Hinton: I don't think we really need to minimize calls to state service, feels nicer to keep the component self-contained as much as possible. Got me thinking a bit on how we'll want to re-use the component for organizations, which might need some refactors but lets handle those down the road.

vincentsalucci (approved):
Resolve merge conflicts then good to go.

vincentsalucci (approved):

PR #1538, opened 2022-03-11
[End User Vault Refresh] Security sub-page

vincentsalucci:
## Type of change

- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Code changes
- **oss-routing.module.ts**: Added new `Security` sub-page and all child routes
- **oss.module.ts**: Added new `SecurityComponent` and `SecurityKeysComponent`
- **account.component.html**: Removed all components that were migrated to new sub-page
- **account.component.ts**: Removed all pieces of code that were migrated to new sub-page
- **change-kdf.component.html**: Added header to the base component - this is only used in one place
- **change-password.component.html**: Added header to the base component - this is only used in one place
- **security-keys.component.html**: Migrated UI components for `KDF` and `API keys`
- **security-keys.component.ts**: Migrated variables/logic for `KDF` and `API Keys`
- **security.component.html**: Created tabs and router outlet for active components
- **security.component.ts**: Migrated variables/logic for `change-password` tab. Added fallback route if the tab is unavailable
- **settings.component.html**: Added `Security` list option. Removed `Two-Factor Login` list option.
- **two-factor-setup.component.html**: Updated header margin with `tabbed-header` class
- **messages.json**: Added new strings
- **base.scss**: Added header related classes
- **navigation.scss**: Added `tabbed-header` class which resembles the existing `org-nav`. Separated header styles and reused `nav-tab` override to match desired style.

## Screenshots
2-security-mp
3-security-2fa
4-security-keys

## Testing requirements
- Changes match the provided Figma wireframes

## Before you submit

### DIFF
--- a/src/app/oss-routing.module.ts
+++ b/src/app/oss-routing.module.ts
@@ -51,6 +51,7 @@ import { SendComponent } from "./send/send.component";
 import { OrganizationGuardService } from "./services/organization-guard.service";
 import { OrganizationTypeGuardService } from "./services/organization-type-guard.service";
 import { AccountComponent } from "./settings/account.component";
+import { ChangePasswordComponent } from "./settings/change-password.component";
 import { CreateOrganizationComponent } from "./settings/create-organization.component";
 import { DomainRulesComponent } from "./settings/domain-rules.component";
 import { EmergencyAccessViewComponent } from "./settings/emergency-access-view.component";
@@ -58,6 +59,8 @@ import { EmergencyAccessComponent } from "./settings/emergency-access.component"
 import { OrganizationsComponent } from "./settings/organizations.component";
 import { PreferencesComponent } from "./settings/preferences.component";
 import { PremiumComponent } from "./settings/premium.component";
+import { SecurityKeysComponent } from "./settings/security-keys.component";
+import { SecurityComponent } from "./settings/security.component";
 import { SettingsComponent } from "./settings/settings.component";
 import { SponsoredFamiliesComponent } from "./settings/sponsored-families.component";
 import { TwoFactorSetupComponent } from "./settings/two-factor-setup.component";
@@ -184,14 +187,32 @@ const routes: Routes = [
             data: { titleId: "preferences" },
           },
           {
+            path: "security",
+            component: SecurityComponent,
+            data: { titleId: "security" },
+            children: [
+              { path: "", pathMatch: "full", redirectTo: "change-password" },
+              {
+                path: "change-password",
+                component: ChangePasswordComponent,
+                data: { titleId: "masterPassword" },
+              },
+              {
+                path: "two-factor",
+                component: TwoFactorSetupComponent,
+                data: { titleId: "twoStepLogin" },
+              },
+              {
+                path: "security-keys",
+                component: SecurityKeysComponent,
+                data: { titleId: "keys" },
+              },
+            ],
+          },
+          {
             path: "domain-rules",
             component: DomainRulesComponent,
             data: { titleId: "domainRules" },
-          },
-          {
-            path: "two-factor",
-            component: TwoFactorSetupComponent,
-            data: { titleId: "twoStepLogin" },
           },
           { path: "premium", component: PremiumComponent, data: { titleId: "goPremium" } },
           { path: "billing", component: UserBillingComponent, data: { titleId: "billing" } },
--- a/src/app/oss-routing.module.ts
+++ b/src/app/oss-routing.module.ts
@@ -51,7 +51,6 @@ import { SendComponent } from "./send/send.component";
 import { OrganizationGuardService } from "./services/organization-guard.service";
 import { OrganizationTypeGuardService } from "./services/organization-type-guard.service";
 import { AccountComponent } from "./settings/account.component";
-import { ChangePasswordComponent } from "./settings/change-password.component";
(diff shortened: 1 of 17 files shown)


src/app/settings/security-keys.component.html
```diff
@@ -0,0 +1,18 @@
...
+<div
+  [ngClass]="{ 'tabbed-header': !showChangeKdf, 'secondary-header': showChangeKdf }"
+  class="border-0 mb-0"
+>
+  <h1>{{ "apiKey" | i18n }}</h1>
+</div>
+<p>
+  {{ "userApiKeyDesc" | i18n }}
+</p>
+<button type="button" class="btn btn-outline-secondary" (click)="viewUserApiKey()">
```
Hinton, line 11: Could you update all buttons to use the button directive from the component library? :)

```suggestion
<button bit-button type="button" (click)="viewUserApiKey()">
```
```

File: src/app/settings/security.component.ts [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+    private router: Router,
+    private route: ActivatedRoute
+  ) {}
+
+  async ngOnInit() {
+    this.showChangePassword = !(await this.keyConnectorService.getUsesKeyConnector());
+    if (!this.showChangePassword) {
+      this.router.navigate(["/settings/security/two-factor"]);
+    }
+  }
```
Hinton, line 24: This could potentially be messy if the user navigates to any security page they will get re-directed to two-factor. Is the goal to hide the change-password page? If so could we move the logic to that page?

src/app/oss-routing.module.ts
```diff
@@ -183,16 +186,34 @@ const routes: Routes = [
...
+                component: TwoFactorSetupComponent,
+                data: { titleId: "twoStepLogin" },
+              },
+              {
+                path: "security-keys",
+                component: SecurityKeysComponent,
+                data: { titleId: "keys" },
+              },
+            ],
+          },
```
Hinton, line 211: This would be a good opportunity to extract these routes from the gigantic routing file similar to #1536

Hinton (changes requested):

src/app/settings/security.component.ts
vincentsalucci: Not sure I follow. The `security` page's default route is `settings/security/change-password` but there is an instance where the `change-password` tab is not available so the component handles navigating to the next available tab. This only happens when the `security` component is initialized via selection from the left-handed vertical menu. `two-factor` is always present and visible. Since the `security` component handles the available tabs, my thought is that is makes the most sense to handle the above scenario here.

src/app/oss-routing.module.ts
vincentsalucci: Neat! Will update.

src/app/settings/security-keys.component.html
vincentsalucci: Absolutely - will go through the modified components and update to the new buttons! Thanks for the reminder.

src/app/oss-routing.module.ts
vincentsalucci: Updated!

src/app/settings/security.component.ts
Hinton: This component is the "layout" component for each security page right? If you say navigate directly to `settings/security/security-keys` It will redirect you to `two-factor` since it triggers the `ngOnInit` for the SecurityComponent.

vincentsalucci: Ah! - thanks so much for clarifying. I've moved the navigation reroute into the `change-password` component which will protect against the scenario above as well as someone trying to route directly to `settings/security/change-password`.

src/app/settings/security-keys.component.html
vincentsalucci: Updated - left one todo regarding translation of bootstraps `btn-sm`

Hinton: @vincentsalucci could you speak with @danielleflinn regarding the `btn-sm` usages? If it's a pattern we want to keep we should add the `size` attribute to `bit-button`.

src/app/settings/change-password.component.html
```diff
@@ -2,7 +2,8 @@
   <h1>{{ "changeMasterPassword" | i18n }}</h1>
 </div>
 
-<app-callout type="warning">{{ "loggedOutWarning" | i18n }}</app-callout>
+<bit-callout type="warning">{{ "loggedOutWarning" | i18n }}</bit-callout>
+<!--TODO Update Component Library with enforcedPolicyOptions callout options-->
```
Hinton, line 6: Should be a separate component, we don't want to pollute the generic callout with policy knowledge.

Hinton (commented):
Could you remove the comments and move those discussions to the Component Library jira tickets? Otherwise we're good.

src/app/settings/change-password.component.html
vincentsalucci: Creating jira task to add this component! Comment removed.

src/app/settings/security-keys.component.html
vincentsalucci: Removed comment and started conversation with design. Will create jira task if needed.

Hinton (approved):
Looks good!

PR #1550, opened 2022-03-16
[SG-34 End User Vault Refresh] Organization Switcher

eliykat:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Add the Organization Picker, using the dropdown menu from the Component Library: https://github.com/bitwarden/jslib/pull/761.

## Code changes

- **src/app/components/organization-picker.component**
  - extract the picker out to its own component
  - apply Tailwind classes instead of Bootstrap utility classes
  - use the new `bit-menu` components
  - take the current org Id as an `@Input` from the parent component (per previous feedback)
- **src/app/organizations/vault/groupings.component.ts** - add org setter which resets the selected grouping when the org changes.
- **src/app/organizations/vault/vault.component.ts** - remove `pipe(first)` because we need to listen to subsequent changes to the orgId to update the active vault.

## Screenshots

Screen Shot 2022-04-10 at 2 47 22 pm

Screen Shot 2022-04-10 at 2 47 31 pm

## Testing requirements

General testing of the new component and switching between orgs from various screens/sub-pages.

## Before you submit


Hinton:
You are probably already thinking of this, but if we implement the "Dropdown menu" for the Component Library we could simplify this menu.

eliykat:
@Hinton yes I'm working on the CL dropdown menu currently, which is a blocker for this PR + #1533

eliykat (commented):
I need to change the language back to Organization Switcher, not sure where I got picker from.

vincentsalucci:
@eliykat This PR should be targeting  `feature/endUserVaultRefresh`.

Hinton:
Tests seems to be failing @eliykat

eliykat:
@Hinton yes, we need some other changes to web to match jslib before builds will pass. Same issue on my other 2 PRs. I believe the blocking PR is https://github.com/bitwarden/web/pull/1595

eliykat:
Have merged a little prematurely with this branch: https://github.com/bitwarden/web/pull/1637
The diff will clean up once that's merged.

eliykat:
Ready for review 🚀

src/app/modules/loose-components.module.ts
```diff
@@ -221,6 +222,7 @@ import { OrganizationBadgeModule } from "./vault/modules/organization-badge/orga
     MasterPasswordPolicyComponent,
     NavbarComponent,
     NestedCheckboxComponent,
+    OrganizationSwitcherComponent,
```
differsthecat, line 225: I think it's best to avoid putting new components in here.
Can this possibly be put in any of the new modules created with other EUVR work? (maybe the `organization-vault` or `vault` modules?)

src/app/components/organization-switcher.component.html
```diff
@@ -0,0 +1,66 @@
...
+      size="45"
+      [circle]="true"
+      [dynamic]="true"
+    ></app-avatar>
+    <div class="org-name tw-ml-3">
+      <span>{{ activeOrganization.name }}</span>
+      <small class="tw-text-muted">{{ "organization" | i18n }}</small>
+    </div>
+    <div class="tw-ml-3">
+      <i class="bwi bwi-angle-down tw-text-main" aria-hidden="true"></i>
```
differsthecat, line 20: The arrow on this switcher looks a little strange to me with how it's centered on the entire switcher itself and floats lower than the org name. Do you think it look better in line with the organization name?

differsthecat:
Looks great! 🎉 
Just a couple of small comments

src/app/modules/loose-components.module.ts
eliykat: Really we need a new `OrganizationsModule` to house all these org components. The `vault` modules are too low level - the org switcher sits outside the vault component and is visible when navigating the other org pages as well. I think this is a refactor we can do after merging all of this, it feels too late in the game to go doing it now. (Ideally all of these "loose components" should be broken out into their own feature modules.)

src/app/components/organization-switcher.component.html
eliykat: I think either would be fine, but I think the Figma agrees with you:
Screen Shot 2022-05-05 at 1 40 58 pm

I'll fix this shortly.

eliykat: Done, here's the updated version:
Screen Shot 2022-05-05 at 2 16 01 pm

eliykat: Update: no truncation required

differsthecat (approved):

PR #1551, opened 2022-03-17
[End User Vault Refresh] Organizations - updated nav and route permissions

eliykat:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

SG-33:
* add top-level Organization page
* enable navigating to the `/organizations` route without specifying an organization id
* update route permissions and templating logic so that ordinary users do not have access to this section anymore

## Code changes

* **src/app/oss-routing.module.ts** - 
  * export the routing permissions for the top-level organization pages. This allows other components to import them and use them when deciding, e.g. whether to display tabs that navigate to those pages. This keeps permissions all in one place and avoids duplication that would lead to them diverging.
  * add generic `/organizations` endpoint
  * update to use new route guards (see below)
* **src/app/services/organization-permissions-guard.service.ts** - this is a refactored version of `OrganizationTypeGuard`. It enforces routing permissions but also brings across some checks from the old `OrganizationGuard` which should apply here as well.
* **src/app/services/organization-redirect-guard.service.ts** - responsible for redirecting the user to a default org where none is specified, or if the user does not have sufficient permissions to access any org, send them back to the home vault page. 
* **src/app/layouts/navbar.component** - add an Organizations link, display it if the user has permissions to access the admin pages
* **src/app/layouts/organization-layout.component.ts** - update to use the same permissions as the routing permissions

Depends on https://github.com/bitwarden/jslib/pull/727.

## Testing requirements

General testing and regression on navigating to Organizations and around Organization tabs with various permissions. (owner, admin, user, custom permissions etc).

## Before you submit


src/app/services/organization-redirect-guard.service.ts
```diff
@@ -0,0 +1,36 @@
...
+import { ActivatedRouteSnapshot, CanActivate, Router } from "@angular/router";
+
+import { I18nService } from "jslib-common/abstractions/i18n.service";
+import { OrganizationService } from "jslib-common/abstractions/organization.service";
+import { PlatformUtilsService } from "jslib-common/abstractions/platformUtils.service";
+import { Permissions } from "jslib-common/enums/permissions";
+import { Utils } from "jslib-common/misc/utils";
+
+@Injectable()
+export class OrganizationRedirectGuardService implements CanActivate {
```
Hinton, line 11: It's not obvious at a glance that this guard checks if any organization matches the conditions. Maybe we should rename it or add a description?

src/app/services/organization-redirect-guard.service.ts
```diff
@@ -0,0 +1,36 @@
...
+    private organizationService: OrganizationService
+  ) {}
+
+  async canActivate(route: ActivatedRouteSnapshot) {
+    const orgs = await this.organizationService.getAll();
+    const permissions = route.data == null ? null : (route.data.permissions as Permissions[]);
+
+    const allowedOrgs = orgs
+      .filter((org) => org.hasAnyPermission(permissions))
+      .sort(Utils.getSortFunction(this.i18nService, "name"));
```
Hinton, line 25: We're reusing the same logic here as in the navbar. What do you think of moving it to say the organization service, with a simpler API?

`organizationService.canViewOrganizationPage()` ... or something.

src/app/oss-routing.module.ts
```diff
@@ -280,18 +302,30 @@ const routes: Routes = [
...
   },
+  {
+    path: "organizations",
+    component: OrganizationLayoutComponent,
+    canActivate: [AuthGuardService, OrganizationRedirectGuardService],
+    data: {
+      permissions: organizationRoutePermissions.all(),
+    },
+    pathMatch: "full",
+  },
```
Hinton, line 313: What do you think of moving the organization routes to it's own router similar to how we've begun splitting it up in the other PRs?

As a way to begin breaking up the application in better modules.

src/app/services/organization-permissions-guard.service.ts
```diff
@@ -4,32 +4,44 @@ import { ActivatedRouteSnapshot, CanActivate, Router } from "@angular/router";
...
+      return this.cancelNavigation();
     }
 
     return true;
   }
+
+  private cancelNavigation() {
+    this.router.navigate(["/"]);
+    return false;
+  }
```
Hinton, line 46: Should we create our own base class for router guards?

Hinton (commented):
Overall looks good, a few optional thoughts I had while reviewing.

src/app/oss-routing.module.ts
eliykat: I think it's a good idea, I've created a separate routing module, I'm now creating a whole separate `organizations.module.ts` as well. You gotta do the PR review though :P

eliykat: Update: I've added the routing module here, will leave the app module for a separate refactor/PR.

src/app/organizations/oss.module.ts
```diff
@@ -0,0 +1,283 @@
+import { DragDropModule } from "@angular/cdk/drag-drop";
```
Hinton, line 1: I believe you forgot to remove this file.

Hinton (changes requested):

src/app/organizations/oss.module.ts
eliykat: Pushing code at 5pm on a Friday 😁

src/app/services/organization-permissions-guard.service.ts
eliykat: PR for review: https://github.com/bitwarden/jslib/pull/732

src/app/services/organization-redirect-guard.service.ts
eliykat: Renamed to `RedirectToOrgAdminGuardService` and added a js-doc comment.
Also raised in Slack whether we can drop `Service` from these names because it's messy.

src/app/services/organization-redirect-guard.service.ts
eliykat: I don't want it in jslib (or the very bloated `OrganizationService`) because it's specific to just one client. I agree with your sentiment though, I've created a `PermissionsService` which is a wrapper around this. This is probably a better interface than dealing with the `organizationRoutePermissions` object directly.

eliykat:
Ready for review again @Hinton

src/app/layouts/navbar.component.html
```diff
@@ -24,6 +24,11 @@
         <li class="nav-item" routerLinkActive="active">
           <a class="nav-link" routerLink="/tools">{{ "tools" | i18n }}</a>
         </li>
+        <ng-container *ngIf="showOrganizations">
```
Hinton, line 27: Do we need a container for these? The `*ngIf` could just be on the `li`.

src/app/layouts/navbar.component.ts
```diff
@@ -1,12 +1,15 @@
...
 
 import { MessagingService } from "jslib-common/abstractions/messaging.service";
+import { OrganizationService } from "jslib-common/abstractions/organization.service";
 import { PlatformUtilsService } from "jslib-common/abstractions/platformUtils.service";
 import { ProviderService } from "jslib-common/abstractions/provider.service";
 import { SyncService } from "jslib-common/abstractions/sync.service";
 import { TokenService } from "jslib-common/abstractions/token.service";
 import { Provider } from "jslib-common/models/domain/provider";
 
+import { NavigationPermissionsService as OrgNavigationPermissionsService } from "../organizations/services/navigation-permissions.service";
```
Hinton, line 11: What do you think of just naming that service `OrganizationNavigationPermissionsService` to begin with, and get rid of these aliasing patterns?

src/app/organizations/guards/permissions.guard.ts
```diff
@@ -1,33 +1,42 @@
...
 import { OrganizationService } from "jslib-common/abstractions/organization.service";
 import { PlatformUtilsService } from "jslib-common/abstractions/platformUtils.service";
+import { Permissions } from "jslib-common/enums/permissions";
 
 @Injectable()
-export class OrganizationGuardService implements CanActivate {
+export class PermissionsGuard extends BaseGuard implements CanActivate {
   constructor(
-    private router: Router,
+    protected router: Router,
```
Hinton, line 13:
```suggestion
    router: Router,
```

src/app/services/services.module.ts
```diff
@@ -45,11 +45,12 @@ import { PasswordRepromptService } from "../../services/passwordReprompt.service
 import { StateService } from "../../services/state.service";
 import { StateMigrationService } from "../../services/stateMigration.service";
 import { WebPlatformUtilsService } from "../../services/webPlatformUtils.service";
+import { PermissionsGuard as OrgPermissionsGuard } from "../organizations/guards/permissions.guard";
+import { NavigationPermissionsService as OrgPermissionsService } from "../organizations/services/navigation-permissions.service";
+import { RedirectToAdminGuard as OrgRedirectToAdminGuard } from "../organizations/services/redirect-to-admin.guard";
```
Hinton, line 50: Same question here, what do you think of just prefixing them with Organization and avoid the aliasing hassle.

src/app/organizations/services/redirect-to-admin.guard.ts
```diff
@@ -0,0 +1,41 @@
...
+import { NavigationPermissionsService } from "./navigation-permissions.service";
+
+/**
+ * Redirects the user to the admin view of the first organization they have access to, or cancels navigation
+ * if they do not have permissions to access admin view for any organization.
+ */
+@Injectable()
+export class RedirectToAdminGuard extends BaseGuard implements CanActivate {
+  constructor(
+    protected router: Router,
```
Hinton, line 19:
```suggestion
    router: Router,
```

src/app/layouts/navbar.component.html
```diff
@@ -24,6 +24,11 @@
...
+          <li class="nav-item" routerLinkActive="active">
+            <a class="nav-link" routerLink="/organizations">{{ "organizations" | i18n }}</a>
```
Hinton, line 29: I've been thinking on this a bit, we could just link directly to the first organization if they only have a single organization. Similar to how the provider link works. Which would let us get rid of the router guard. No real preference though, but would be nice to update the provider to use the same pattern if we switch?

src/app/layouts/navbar.component.ts
eliykat: I also thought about this, BUT all aliasing patterns will be dropped once I introduce the `OrganizationModule`. Most things are used within that module and won't need to be prefixed with `Org`. So this is just temporary.

src/app/services/services.module.ts
eliykat: Same as above

src/app/organizations/guards/permissions.guard.ts
eliykat: ty 🙈

src/app/layouts/navbar.component.html
eliykat: The nice thing about the router guard is that it lets the navbar direct to a route without having to know where it goes. But, it does duplicate this logic a little, and it's a bit counterintuitive having a `canActivate` guard which always returns false. I'll bring it in line with the providers.

eliykat:
@Hinton can you please take a look at the routing on this? I can't navigate to `organizations/:organizationId` and have it redirect to the org vault page. Instead, it loads the org layout in the *top level* router and doesn't trigger the redirect. But, if I click on one of the tabs, it loads correctly. I've played with the nesting/routes a lot but none of it's made any difference.

https://user-images.githubusercontent.com/31796059/159597763-5d1182b0-f509-44ec-ad59-c0dda90bef35.mov

src/app/layouts/navbar.component.html
eliykat: This is done

eliykat:
Feedback has been actioned, also cherry-picked https://github.com/bitwarden/web/pull/1560 just so we could have the fix in this branch now.

src/app/organizations/organization-routing.module.ts
```diff
@@ -0,0 +1,217 @@
...
+import { ImportComponent } from "./tools/import.component";
+import { InactiveTwoFactorReportComponent } from "./tools/inactive-two-factor-report.component";
+import { ReusedPasswordsReportComponent } from "./tools/reused-passwords-report.component";
+import { ToolsComponent } from "./tools/tools.component";
+import { UnsecuredWebsitesReportComponent } from "./tools/unsecured-websites-report.component";
+import { WeakPasswordsReportComponent } from "./tools/weak-passwords-report.component";
+import { VaultComponent } from "./vault/vault.component";
+
+const routes: Routes = [
+  {
```
Hinton, line 32: Do we have a route to the organizations list page?

src/app/layouts/navbar.component.html
```diff
@@ -24,6 +24,11 @@
         <li class="nav-item" routerLinkActive="active">
           <a class="nav-link" routerLink="/tools">{{ "tools" | i18n }}</a>
         </li>
+        <li *ngIf="organizations.length >= 1" class="nav-item" routerLinkActive="active">
+          <a class="nav-link" [routerLink]="['/organizations', organizations[0].id]">{{
+            "organizations" | i18n
+          }}</a>
```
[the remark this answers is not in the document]
eliykat: That's what the organization picker is for (still WIP). This link only needs to send them to the first org in their list, then they user the picker from that page to change orgs if they need to.

src/app/organizations/organization-routing.module.ts
eliykat: Same as above

Hinton (approved):

PR #1558, opened 2022-03-21
Update configurations for self-hosted

MGibson1:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
Related to bitwarden/server#1921

Sets up configurations for a locally-run, self-hosted development instance.

Changes the dev settings to a mutually exclusive set of settings for cloud-like and self-hosted like. The selection is dependent on env variables when run

## Code changes
* **base.json**: Set default port for dev cloud
* **development.json**: split dev object into `cloud` and `self hosted`, updating proxies to point to respective services.
* **package.json**: build and watch scripts for a self-hosted development instance
* **webPlatformUtils**: return self-hosted true for dev self host environments as well
* **webpack.config**: 
  * set correct dev environment depending on whether self_hosted is set. 
  * Make `SELF_HOSTED` available to webPlatformUtils. 
  * Default port to 8080

#### Testing requirements

None, dev only

## Before you submit


webpack.config.js
```diff
@@ -142,6 +144,7 @@ const plugins = [
   new webpack.EnvironmentPlugin({
     ENV: ENV,
     NODE_ENV: NODE_ENV === "production" ? "production" : "development",
+    SELF_HOSTED: SELF_HOSTED,
```
Hinton, line 147: This is a bit messy, do we need this?

I believe @joseph-flinn and I settled on using `NODE_ENV` to switch between development and production code parts, and `ENV` to switch between cloud, qa, self-hosted.

webpack.config.js
```diff
@@ -15,8 +15,10 @@ const pjson = require("./package.json");
 
 const ENV = process.env.ENV == null ? "development" : process.env.ENV;
 const NODE_ENV = process.env.NODE_ENV == null ? "development" : process.env.NODE_ENV;
+const SELF_HOSTED = ENV === "selfhosted" || process.env.SELF_HOSTED === "true";
 
 const envConfig = config.load(ENV);
+envConfig.dev = SELF_HOSTED ? envConfig.dev.selfHosted : envConfig.dev.cloud;
```
Hinton, line 21: We have a config file for self hosted which we could put this in. We need to be careful though since the ports don't match what users would typically use. Unless we update all self-hosted instances to use the new ports which would let us simplify things?

webpack.config.js
MGibson1: we settled on `NODE_ENV` signifying isDev and `ENV` indicating the type of environment. This is a fourth environment.

We either need `ENV=development` and `ENV=self_hosted_development` to launch dev servers OR we need another env variable to signify that we're developing for self hosted. I like this way better than the three-way ENV switch

webpack.config.js
Hinton: I wonder if we should just bite the bullet and add support for `env.development.json`.

webpack.config.js
MGibson1: It would have run as if it were cloud production, but it's all torn out now.

PR #1579, opened 2022-04-04
[EC-143] [BEEEP] Allow linking to ciphers

Hinton:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Adds support for for linking directly to ciphers.

As a part of this effort I re-factored our different redirection flows into a single flow. This expands the regression testing needed but simplifies the flow for the future.

EC-143

Depends on https://github.com/bitwarden/jslib/pull/760

## Code changes

- **src/app/accounts/lock.component.ts:** We previously used `navigate` which discards the query params. `navigateByUrl` keeps the query params which keeps the vault filters and cipherId.
- **src/app/accounts/login.component.ts**: Changed the old `stateService.setLoginRedirect` flow to use `routerService.setPreviousUrl`. Redirects to whichever page is set as `loginRedirect` after login.
- **src/app/accounts/register.component.ts**: Migrated from `stateService.setLoginRedirect` to `routerService.setPreviousUrl`.
- **src/app/app.component.ts**: Set previousUrl on auth blocked.
- **src/app/common/base.accept.component.ts**: Remove accept logic where we set redirect as it's now handled by the previous url logic.
- **src/app/guards/home.guard.ts**: Add Home Guard which controls how `/` is redirected.
- **src/app/organizations/vault/vault.component.ts**, **src/app/vault/vault.component.ts**: Add `cipherId` to parms which is the opened cipher. This required some extra logic to remove once the modal was closed, and to ensure selecting a filter didn't discard it.
- **src/app/vault/ciphers.component.html**: Update the link to be a proper link instead of `#`.
- **src/app/services/router.service.ts**: Simplify logic as the previous logic was very difficult to understand. Now utilizes the fact that the last child has a combined state.

## Testing requirements

We need to regression test the following areas:
* Accept Organization & Emergency Access
* The `?premium` and `?org` flags for login/register.
* General logging in, registering and flows.

All of the above flows are typically tricky areas and we should spend some time making sure everything works as we expect it.

## Before you submit


src/app/services/organization-guard.service.ts
```diff
@@ -4,30 +4,35 @@ import { ActivatedRouteSnapshot, CanActivate, Router } from "@angular/router";
...
-    private organizationService: OrganizationService
+    private organizationService: OrganizationService,
+    private syncService: SyncService
   ) {}
 
   async canActivate(route: ActivatedRouteSnapshot) {
+    // TODO: We need to fix this issue once and for all.
+    if ((await this.syncService.getLastSync()) == null) {
+      await this.syncService.fullSync(false);
+    }
```
Hinton, line 23: I think the correct way to resolve this would be to perform the sync separately before we allow the user into the vault.

src/app/app.component.ts
```diff
@@ -91,11 +91,17 @@ export class AppComponent implements OnDestroy, OnInit {
...
+            break;
           case "loggedOut":
+            this.routerService.setPreviousUrl(null);
+            this.notificationsService.updateConnection(false);
+            break;
           case "unlocked":
             this.notificationsService.updateConnection(false);
             break;
           case "authBlocked":
+            window.setTimeout(() => this.routerService.setPreviousUrl(message.url), 500);
```
MGibson1, line 104: Did this just feel like a reasonable timeout, or are we trying to slow things down intentionally with this? I believe we just need this to finish prior to `/` submitting, right?

MGibson1 (commented):
I like how this does a lot of work to elevate the url and history to a keeper of state.

I do have a few questions, though

src/app/accounts/login.component.ts
```diff
@@ -145,10 +146,9 @@ export class LoginComponent extends BaseLoginComponent {
       }
     }
 
-    const loginRedirect = await this.stateService.getLoginRedirect();
-    if (loginRedirect != null) {
-      this.router.navigate([loginRedirect.route], { queryParams: loginRedirect.qParams });
-      await this.stateService.setLoginRedirect(null);
+    const previousUrl = this.routerService.getPreviousUrl();
+    if (previousUrl) {
```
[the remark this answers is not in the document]
Hinton: The unauthenticated pages do not save history by default https://github.com/bitwarden/web/pull/1579/files#diff-0e6ec801f8b1eec3e9ac4d550e94707bdf7899a06714478fdcee2cb948d0108dR77

src/app/app.component.ts
Hinton: We have the same timeout a couple of lines below. I couldn't be bothered to figure out why. 😿

Hinton:
Note for myself:

* Remove links to `/` for accept pages.
* Can we remove timeout in `app.component`.

Hinton:
Seemed to work without the timeouts during my testing.

src/app/accounts/two-factor.component.html
```diff
@@ -138,7 +138,7 @@
                 aria-hidden="true"
               ></i>
             </button>
-            <a routerLink="/" class="btn btn-outline-secondary btn-block ml-2 mt-0">
+            <a routerLink="/login" class="btn btn-outline-secondary btn-block ml-2 mt-0">
```
MGibson1, line 141: Commenting on this for documentation. You shouldn't link to `/` because that landing location is now the `HomeGuard`, which always redirects. You can get stuck in a redirect loop where angular refuses to continue loading.

PR #1576, opened 2022-03-31
[euvr] Subscription/Billing updates

vincentsalucci:
## Type of change

- [ ] Bug fix
- [X] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other

## Code changes
- **organization-billing.component.html**: Copied over existing `user-billing.component.html` to maintain backwards compatibility for organizational settings (to be updated in the near future)
- **organization-billing.component.ts**: Removed base class and gave this component all the capabilities necessary to maintain backwards compatibility for organizational settings (to be updated in the near future)
- **oss-routing.module.ts**: Updated routing to match new design pattern
- **oss.module.ts**: Added newly created components
- **payment-method.component.html**: Created. Took useful pieces from the `user-billing.component`
- **payment-method.component.ts**: Created. Took useful pieces from the `user-billing.component`
- **premium.component.html**: Updated to match new design. Updated with new `CL` components where applicable
- **premium.component.ts**: Updated route to redirect when a user already has premium
- **settings.component.html**: Removed old navigation items and replaced with single `Subscription` option
- **settings.component.ts**: Added additional logic that will handle hiding the `Subscription` option under certain conditions.
- **subscription.component.html**: Created container for routing to selected tabs within the `Subscription` navigation option
- **subscription.component.ts**: Created. Handles supply of any logic needed for adjusting routes
- **subscription.module.ts**: Created. Handles lazy loading of child routes for the `Subscription` navigation option
- **user-billing.component.html**: Removed all `payment-method` related UI and updated with new `CL` components where applicable
- **user-billing..component.ts**: Removed all `payment-method` variables/logic/methods
- **user-subscription.component.html**: Adjusted for updated designs.  uUpdated with new `CL` components where applicable
- **user-subscription.component.ts**: Added helper method for title change. Updated navigation scheme on redirect.
- **messages.json**: Added `billingHistory`

## Screenshots
**Hidden Subscription due to proper conditions**
4-security-keys
5-sub-sub
6-sub-payment
7-sub-billing
8-sub-self-host

## Testing requirements
- Subscription flows for both cloud and self-hosted
- Organizational billing (should work as it currently does - need regression)

## Dependency
- [jslib - Separate Billing API](https://github.com/bitwarden/jslib/pull/750)

## Before you submit
- [X] I have checked for **linting** errors (`npm run lint`) (required)
[...]

PR #1595, opened 2022-04-13
[End User Vault Refresh] SG-16: Organization filters

differsthecat:
https://bitwarden.atlassian.net/browse/SG-16

## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
The product initiative here is to implement the filters section of the end user vault refresh epic. This includes:

1. Allowing top level filters (types, folders, collections) to be collapsable
2. Adding a filter for organizations (or lack there of)
3. Improved accessibility (using correct tags, expanded attributes, etc.)
4. Limiting vault search to the active filter selected

There are some other initiatives in here, mostly around refactoring. These include:

1. Splitting the GroupingsComponent into a module with several smaller components. This is also mostly done in jslib, and is being implemented in this PR.
2. Created a SharedModule as an effort to clean up AppModule, and to reuse shared modules in other modules. This is a pretty standard practice in Angular applications.
3. Creating a `VaultModule` with `IndividualVaultModule` and `OrganizationVaultModule` children that share a service. This is to share search field placeholder text logic between the org and individual vault components. There is certainly a lot more that could be shared between these two and refactored, but this already seemed like scope creep and only went far enough to achieve the need for filters.
4. Extracting the scss from the filters sidebar into a separate file. This was mostly a necessity in order to make sense of what was going on with the styles during development, but is an extra refactor none the less.

## Code changes
- Split the `GroupingsComponent` into a `VaultFiltersModule` with components for each filter type
- Developed a new filter type for organizations - `OrganizationFilterComponent`
- Allowed for filter headers to be collapsible as individual nodes (most of this work is in jslib)
- Massive HTML restructuring of filters in order to get the accessibility changes in. Previously each node consisted of an `<a>` tag with a `click` event that performed the filter, but that `<a>` tag could have any number of `<span>` or `<i>` tags inside of it that had their own click events. I changed all of this so every clickable element is a `<button>`, and added the attributes requested by design.
- Massive SCSS restructuring to match the HTML restructuring. 
- Created a `VaultModule` with `IndividualVaultModule` and `OrganizationVaultModule` children that share a service.
- Created a `SharedModule` for passing shared code around to feature modules
- Created a `LooseComponentsModule` as a temporary space for homeless components to exist and be consumed by feature modules.

## Screenshots

Screen Shot 2022-04-13 at 3 23 59 PM

## Before you submit


src/app/settings/user-billing.component.ts
```diff
@@ -47,7 +47,9 @@ export class UserBillingComponent implements OnInit {
     if (this.organizationId != null) {
       this.billing = await this.apiService.getOrganizationBilling(this.organizationId);
     } else {
-      this.billing = await this.apiService.getUserBilling();
+      // let history = await this.apiService.getUserBillingHistory();
+      // let payment = await this.apiService.getUserBillingPayment();
+      this.billing = new BillingResponse(null);
```
addisonbeck, line 52: What's this?

src/app/vault/vault-filter/vault-filter.module.ts
```diff
@@ -0,0 +1,48 @@
+import { NgModule } from "@angular/core";
```
addisonbeck, line 1: I would like to move this module into a new folder at `src/app/modules/vault-filter`. This is similar to what I've got right now on the desktop [PR](src/app/modules/vault-filter/vault-filter.module.ts). This is also something I can do next week. Any thoughts?

src/app/settings/user-billing.component.ts
differsthecat: I didn't mean to commit that actually, it was a work around to get the latest jslib in. [This PR ](https://github.com/bitwarden/web/pull/1576) has the real changes we want to have

src/app/organizations/vault/vault-filter/vault-filter.component.ts
```diff
@@ -0,0 +1,27 @@
...
+import { FolderService } from "jslib-common/abstractions/folder.service";
+import { I18nService } from "jslib-common/abstractions/i18n.service";
+import { StateService } from "jslib-common/abstractions/state.service";
+import { CollectionData } from "jslib-common/models/data/collectionData";
+import { Collection } from "jslib-common/models/domain/collection";
+import { Organization } from "jslib-common/models/domain/organization";
+import { CollectionDetailsResponse } from "jslib-common/models/response/collectionResponse";
+import { CollectionView } from "jslib-common/models/view/collectionView";
+
+import { VaultFilterComponent as BaseVaultFilterComponent } from "../../../vault/vault-filter/vault-filter.component";
```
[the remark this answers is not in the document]
differsthecat: Yeah, that sounds great! I think there's a way stronger case for it here in web vs what it was in Browser

src/app/vault/vault-filter/vault-filter.module.ts
differsthecat: Yep, that's fine! It made sense to me here in the vault context, but I think it works well there too

package-lock.json
```diff
@@ -88,7 +88,7 @@
       }
     },
     "jslib/angular": {
-      "name": "@bitwarden/jslib-common",
+      "name": "@bitwarden/jslib-angular",
```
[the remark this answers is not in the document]
differsthecat: I thought that it is good practice to do commit these changes, but I can undo them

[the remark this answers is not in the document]
addisonbeck: No worries, I'll let someone else chime in there later if need be. I don't know what the standard is :)

src/app/vault/vault-filter/vault-filter.module.ts
addisonbeck: Cool. I don't disagree that it works where it is, I just think moving away from having business verbiage as top level structures is in our best interests long term.

src/app/organizations/vault/vault-filter/vault-filter.component.ts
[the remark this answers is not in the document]
addisonbeck: I think I was able to just delete the extra component completely and account for both cases in `vault-filter.component`

src/app/vault/vault-filter/vault-filter.module.ts
addisonbeck: Done

src/app/settings/user-billing.component.ts
addisonbeck: I think we will need to leave this in, merge it, and fix it after the linked PR is merged. We are both dependent on each other to get web to build right now 😵‍💫

addisonbeck:
I am aware of the build issue here. Devops has a potential fix, but we think this is an anomaly with this branch that may self correct. I don't think it should hold up merging, considering the target branch doesn't build anyway :)

Hinton:
@addisonbeck The reason it doesn't build is due to there being a package in `package-lock` that gets put in jslib. This often happens when someone bumps jslib packages and forget to refresh the client. (It should go away with the move to monorepo)

src/app/modules/loose-components.module.ts
```diff
@@ -0,0 +1,483 @@
...
+    UserSubscriptionComponent,
+    VaultTimeoutInputComponent,
+    VerifyEmailComponent,
+    VerifyEmailTokenComponent,
+    VerifyMasterPasswordComponent,
+    VerifyRecoverDeleteComponent,
+    WeakPasswordsReportComponent,
+  ],
+})
+export class LooseComponentsModule {}
```
differsthecat, line 483: I'm not a big fan of this module name but I'm struggling to think of an alternative.
`MiscellaneousComponentsModule` feels a _little_ better to me

addisonbeck: I want the name to give the impression that this is a deprecated way of doing things, and in my opinion putting "Misc" on classes usually encourages people to use them more than it discourages.

src/app/modules/vault/vault.module.ts
```diff
@@ -0,0 +1,19 @@
...
+  imports: [SharedModule, VaultFilterModule, LooseComponentsModule],
+  exports: [SharedModule, VaultFilterModule, LooseComponentsModule],
+  providers: [
+    {
+      provide: VaultService,
+      useClass: VaultService,
+    },
+  ],
+})
+export class VaultModule {}
```
Hinton, line 19: Feels like this should be a module under `organizations`, and not vault.

src/app/modules/vault/modules/organization-vault/organization-vault.component.ts
```diff
@@ -127,6 +128,8 @@ export class VaultComponent implements OnInit, OnDestroy {
     this.ciphersComponent.showAddNew = vaultFilter.status !== "trash";
     this.activeFilter = vaultFilter;
     await this.ciphersComponent.reload(this.buildFilter(), vaultFilter.status === "trash");
+    this.vaultFilterComponent.searchPlaceholder =
+      this.vaultService.calculateSearchBarLocalizationString(this.activeFilter);
     this.go();
     this.go();
```
differsthecat, line 134: Looks like `this.go()`  is duplicated here (I think I did this in the initial draft🙈 )

src/app/modules/loose-components.module.ts
differsthecat: That's a good point!

Hinton:
Found the issue, seems jslib/common wanted node-forge 1.3.1, and web for some reason resolved it to 1.3.0. I hard coded web to use the latest for now, the move to mono repo should resolve this permanently.

src/app/modules/vault/modules/organization-vault/organization-vault.component.ts
addisonbeck: [[bug] Remove duplicate this.go() call](https://github.com/bitwarden/web/pull/1595/commits/3a29ead2e5f33b0b7413627e6afcf2f996427f40)

src/app/modules/vault/vault.module.ts
addisonbeck: Assuming you mean you'd want `OrganizationVaultModule` to be a `VaultModule` underneath an `OrganizationModule` or something like that:

I'm open to discussing this, because I understand that the current components are organized this way and I am sure there is a good reason for it, but in my opinion this is a flaw in the application structure. These Vault components have *WAY* more in common with each other than they do with the larger concepts of organizations, or lack there of. They should be near each other in the file system to more easily distinguish that similarity and to have access to shared code without deep relative imports to far away areas of the file structure. 

The `OrganizationsModule` (theoretically, since this doesn't exists) would still only import the `OrganizationVaultModule`, the `VaultModule` itself is only intended to be consumed by its children and does not declare or export any of its children.

PR #1614, opened 2022-04-25
[EC-151] Hide Subscription/Billing information for Provider-managed organizations

eliykat:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Hide billing/subscription information for client organizations managed by Providers.

## Code changes

- **src/app/organizations/settings/organization-subscription.component** - display splash screen if user can't manage billing. Note that we toggle between light/dark mode images based on the theme (the Tailwind `dark:` modifier) setting the CSS `content` property. This is what we do for the logo in our traditional stylesheet.

- **src/app/organizations/settings/settings.component** - hide the Billing menu item if the user can't manage billing

- **src/app/oss-routing.module.ts**  and  **src/app/services/organization-type-guard.service.ts-** add the permission check to the billing route guard as well

- **src/app/organizations/settings/account.component** - disable the Billing Email and Business Name fields if user can't manage billing. (Business name is part of billing data)

- **tailwind.config.js** 
  - we currently do not use the Tailwind dark mode properly, so the `dark` modifier doesn't work. This change means that Tailwind will recognise `tw-light` and `tw-dark` classes and apply modifiers accordingly. Unfortunately, this name is not configurable and so we have to double up on our theming classes everywhere (using both `theme_dark` and `tw-dark` for example). 
  - all other changes are updating the theming logic to ensure these classes are added and removed consistently
  - this might need some related refactors for consistency, e.g. changing `tw-theme.css` to use the `tw-` prefix, updating storybook templates, maybe even create a `ThemeService` to abstract some of the manual theme setting logic (and move it out of `PlatformUtilsService`). Feedback is welcome on how to manage this.

Depends on:
* https://github.com/bitwarden/jslib/pull/777

## Screenshots

Screen Shot 2022-04-25 at 10 55 46 am
Screen Shot 2022-04-25 at 10 55 07 am

## Testing requirements

* as an Owner of an organization managed by a Provider,
  * I should get the splash screen on the Subscriptions page, and no other info
  * I should not be able to access the Billing page
  * I should not be able to change the Business Name or Billing Email from the My Organization page
* as a Provider user, there should be no change to my ability to access or change anything on these pages (regression testing)

## Before you submit


tailwind.config.js
```diff
@@ -1,4 +1,7 @@
 /* eslint-disable no-undef, @typescript-eslint/no-var-requires */
 const config = require("./jslib/components/tailwind.config.base");
 
-module.exports = config;
+module.exports = {
+  ...config,
+  darkMode: "class",
+};
```
Hinton, line 7: We unfortunately can't use tailwind dark since we need to support multiple themes on some clients. @danielleflinn and my approach has been to use css variables for colors.

Hinton (changes requested):

tailwind.config.js
Hinton: For svg icons we used `grayscale` modifier, and `currentColor` to tweak things.

Hinton: One possibility is to make an svg template per https://angular.io/guide/svg-in-templates

eliykat: I went with the SVG template approach. The colours do not seem to align particularly with our theming, so I hard-coded them into the component rather than trying to export them from the css. There's also a bit of logic to handle theme changes, ideally this would be abstracted but the `platformUtilsService.onDefaultSystemThemeChange` method doesn't provide any way to deregister the event handler. I should create a `ThemeService` to pull this out of `PlatformUtilsService` and provide this functionality, but I haven't done that at this stage.

eliykat: https://user-images.githubusercontent.com/31796059/165219798-a30c03b1-0348-4898-a5fb-cd8bd36b5219.mov

src/app/organizations/settings/subscription-hidden.component.ts
```diff
@@ -0,0 +1,55 @@
...
+    [ThemeType.Light]: {
+      shape: "#E1E5EE",
+      line: "#2F343D",
+    },
+    [ThemeType.Dark]: {
+      shape: "#2F343D",
+      line: "#E1E5EE",
+    },
+  };
+  colors = this.themes[ThemeType.Light];
```
Hinton, line 23: Could we use css variables here?

tailwind.config.js
Hinton: For context to my comment below. We would like to avoid hard-coding color variants since it makes looking them up fragmented, and will making porting this component to other clients a chore.

src/app/organizations/settings/subscription-hidden.component.ts
eliykat: Yes, this works! I'll raise it with design to ensure we can use standardised colours.

tailwind.config.js
eliykat: Discussion continued in your comment below

eliykat:
@Hinton This is ready to review again. Design provided a simplified image which aligns with our colour scheme. Unfortunately there are still 2 colors, because the "three people" can't be transparent, otherwise they look like this:

Screen Shot 2022-05-06 at 2 07 55 pm

So we need 2 colors (fill and stroke), so we can't just use CSS `currentcolor`. Instead, I've just included it in the component and used CSS variables to match the color scheme. I think that's still an improvement.

https://user-images.githubusercontent.com/31796059/167065474-7ddbe4f0-76e0-457d-ba5e-07ab3da7a703.mov

src/locales/en/messages.json
```diff
@@ -4895,6 +4895,18 @@
...
+    "message": "Managed by $PROVIDER$",
+    "placeholders": {
+      "provider": {
+        "content": "$1",
+        "example": "Managed Services Company"
+      }
+    }
+  },
+  "billingContactProviderForAssistance": {
+    "message": "Please reach out to them for further assistance"
```
Hinton, line 4908: Should we add a description here?

src/app/organizations/settings/organization-subscription.component.ts
```diff
@@ -11,6 +12,30 @@ import { PlanType } from "jslib-common/enums/planType";
 import { Organization } from "jslib-common/models/domain/organization";
 import { OrganizationSubscriptionResponse } from "jslib-common/models/response/organizationSubscriptionResponse";
 
+// Include the SVG here so that it forms part of the DOM and can use CSS color vars
+const billingHiddenSvg = `
+<svg width="216" height="231" viewBox="0 0 216 231" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Hinton, line 17: Lets move the CSS into the template itself.

src/locales/en/messages.json
eliykat: Done

src/app/organizations/settings/organization-subscription.component.ts
eliykat: Yes, I think it's just tidiest to put it in its own component, even though it feels heavyweight.

I tweaked the SVG to have a consistent `mask` property for both themes, it looked fine in Firefox but I can see not in Chrome. Will go back to design to fix it.

eliykat:
I think this should wait until #1640 is merged - I'd rather resolve merge conflicts on this PR than delay all the EUVR changes.

src/app/organizations/settings/subscription-hidden.component.ts
```diff
@@ -0,0 +1,8 @@
+import { Component } from "@angular/core";
+
+// Component is used so that the SVG can embed CSS color variables
+@Component({
+  selector: "app-org-subscription-hidden",
```
Hinton, line 5: Thoughts on standardizing some form of naming convention for svg? It's a bit unclear looking at the parent component that this is only an icon.

Hinton: `app-icon-subcription-hidden` maybe?

eliykat: Good idea, but I'd prefer `image` as the more generic descriptor.

So that would be
`app-image-org-subscription-hidden` (selector)
`ImageSubscriptionHiddenComponent` (class name)
`image-subscription-hidden.component` (file name)

?

Not sure if there's a clear advantage between prefix and suffix, prefix has the advantage of collecting them together in the folder structure.

Hinton: 🤷 either works. I mostly want to easily identify that it's an image.

eliykat: This is done

eliykat:
* addressed feedback above
* resolved conflicts after EUVR merged to master
* 1 additional change required in jslib as a result of EUVR: https://github.com/bitwarden/jslib/pull/798

Hinton (approved):

PR #1623, opened 2022-04-27
[SG-32] Add Ownership badge to vault items

differsthecat:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

[Context](https://bitwarden.atlassian.net/browse/SG-32)
Add a badge that shows the owner of vault items for users that belong to organizations.

## Code changes

- Added **organization-name-badge.component.ts:**  Component that uses `bit-badge` from the component library, with some additional logic to use custom colors. I made a decision here to use the Bitwarden blue for the `Me` badge to keep consistency, for example in English it calculates to a not so nice red.

- Added **get-organization-name.pipe.ts:** Added this pipe in order to avoid using a method call in the ciphers template. Using method calls in templates can degrade performance a ton because they are called for every change detection cycle. With a pure pipe like this, it is only called if any input changes. This may be a bit of pre-optimization, but I deemed it important in this component where there could be tons of rows

- Modified **ciphers.component.html:** Removed collection icon and added the badge as a row. It only shows up if the user belongs to an organization and if it is in the individual vault view.

## Screenshots

Demo:

https://user-images.githubusercontent.com/8926729/165625495-2d12e3c6-beef-419b-9c80-9fbcb0afb6f0.mov

## Before you submit


src/app/modules/vault-filter/components/organization-name-badge.component.html
```diff
@@ -0,0 +1,9 @@
+<button
+  bit-badge
+  [style.color]="textColor"
```
addisonbeck, line 3: Just wondering - did you try creating a component scoped stylesheet before doing this inline?

I've tried in the past, and it never gets included in compilation and I have no clue why 😵‍💫

src/app/modules/vault-filter/components/organization-name-badge.component.ts
```diff
@@ -0,0 +1,62 @@
...
+      } else {
+        return Math.pow((c + 0.055) / 1.055, 2.4);
+      }
+    });
+
+    const L = 0.2126 * c[0] + 0.7152 * c[1] + 0.0722 * c[2];
+    return L > 0.179 ? "black !important" : "white !important";
+  }
+
+  emitonOrganizationClicked() {
```
addisonbeck, line 59: emitOn?

src/app/modules/vault-filter/vault-filter.module.ts
```diff
@@ -28,8 +30,10 @@ import { VaultFilterComponent } from "./vault-filter.component";
     OrganizationOptionsComponent,
     StatusFilterComponent,
     TypeFilterComponent,
+    GetOrgNameFromIdPipe,
+    OrganizationNameBadgeComponent,
```
addisonbeck, line 34: I'm not sure `VaultFiltersModule` is the right declarer for this. `OrganizationNameBadgeComponent`  isn't used by the anything else in `VaultFiltersModules`, it's used by `IndividualVaultModule`. Do you think that make `IndividualVaultModule` a more fitting home?

If it is placed here for code reuse, I suppose that makes sense, but even then the fact that this component is used for a filter is dictated by the `IndividualVaultComponent` - nothing in the component references a filter. If reuse is a concern, perhaps this get its own module.

addisonbeck (commented):
Great work!

src/app/modules/vault-filter/vault-filter.module.ts
differsthecat: That's a good point, `IndividualVaultModule` is in fact the better place for this as it's not present if in the organization view. I'll test it and see if it works in there. I have concerns that there may be errors on the org vault page, but if Angular evaluates the ngIf before it looks for the component, it should be fine to move.

src/app/modules/vault-filter/components/organization-name-badge.component.html
differsthecat: Maybe it can be one with ngClass 🤔

src/app/modules/vault-filter/vault-filter.module.ts
addisonbeck: Okay. I think if that doesn't work there is no issue with creating an `OrganizationNameBadgeModule` (or just a `BadgeModule` if this can be abstracted in such a way). It might seem pedantic, but a reused component is a reused component.

src/scss/buttons.scss
```diff
@@ -248,3 +248,7 @@ button.no-btn {
     color: themed("textSuccessColor");
   }
 }
+
+.org-name-badge {
+  width: 100px;
+}
```
Hinton, line 254: Please use tailwind for this, `tw-w-24`. We shouldn't be writing any new css at all. https://tailwindcss.com/docs/width

src/app/vault/ciphers.component.html
```diff
@@ -54,6 +45,13 @@
           <br />
           <small appStopProp>{{ c.subTitle }}</small>
         </td>
+        <td *ngIf="organizations.length > 0 && !organization" class="org-name-badge">
```
Hinton, line 48: Can we just leave out a `td`? Won't that cause the option button to occupy this slot?

src/app/modules/vault-filter/pipes/get-organization-name.pipe.ts
```diff
@@ -0,0 +1,20 @@
...
+  pure: true,
+})
+export class GetOrgNameFromIdPipe implements PipeTransform {
+  constructor(private i18nService: I18nService) {}
+
+  transform(value: string, organizations: Organization[]) {
+    let orgName = organizations.find((o) => o.id === value)?.name;
+    if (orgName == null) {
+      orgName = this.i18nService.t("me");
+    }
```
Hinton, line 17: I think this condition makes the pipe very limited in re-use. And would suggest moving the logic into the organization name badge component instead.

Hinton (dismissed):

differsthecat:
> What's up with all the commits in this PR? I assume it only added the last?

@Hinton Yeah it's a little strange now, but I needed the new web filters changes to base these off of but that PR isn't merged yet. When it gets merged, I'll change the base for this one and those should fall away 🙂

src/app/vault/ciphers.component.html
differsthecat: Absolutely, I put it in a `td` with the community PR we have in mind, but it can easily be put into a column when that work is resumed.

Hinton: Sorry for being a bit unclear, my worry is that if we omit the `td` element for some items it will cause things to shuffle around and occupy the wrong slots. Or do we always display this `td`?

differsthecat: Ah I understand now! I can put the `ngIf` on the `app-org-badge` itself, then the `td` will always be there.

differsthecat: Looked into this a bit, when the `td` for the org badge doesn't render, the one after it (the options dropdown) still flows to the right most side of the screen, and the td before it (clickable cipher name) occupies the empty space.
If we move the check and always render the td, then there will be an empty bit of space where the badge would go in cases where it doesn't show.

I can move the ngIf so that we always render the `td`, but right now with the current state of the vault not having headers, it appears fine to leave how it is.

src/app/modules/vault-filter/vault-filter.module.ts
differsthecat: The issue I was running into ended up being a circular module import, because I need to import this into `LooseComponentsModule` to use in the CiphersComponent.
Splitting OrganizationNameBadgeComponent into its own module did the trick!

src/app/modules/vault-filter/pipes/get-organization-name.pipe.ts
differsthecat: Moved out of the pipe and into the badge component!

addisonbeck (approved):

PR #1685, opened 2022-05-16
Allow all hosts for dev environment

MGibson1:
This is needed to enable testing of hCaptcha through a host proxy.

## Type of change

- [ ] Bug fix
- [ ] New feature development
- [x] Other

## Objective

hCaptcha does not allow execution from localhost (https://docs.hcaptcha.com/#local-development), to remediate, the suggested solution is to make up a non-localhost uri and specify it as a known host on your system. That way we get around these rules. Unfortunately, we now hit incorrect host header issues with our webpack server.

The solution to this new problem is to allow all hosts headers as done here. Given that this is for the development configuration, there should be no risk
## Before you submit

### DIFF
--- a/config/development.json
+++ b/config/development.json
@@ -3,7 +3,7 @@
     "notifications": "http://localhost:61840"
   },
   "dev": {
-    "allowedHosts": "all",
+    "allowedHosts": "bitwarden.local",
     "proxyApi": "http://localhost:4000",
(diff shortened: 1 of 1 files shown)


jlf0dev (dismissed):

MGibson1:
Documentation updates and `local.json` configs are more appropriate. I will add an `hCaptcha` testing page once our public documentation is editable again

PR #1698, opened 2022-05-19
[PS-616] Feature/require auth for enroll password reset

MGibson1:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Require user secret validation for password reset enrollment.

## Code changes
* **enroll-master-password-reset.component**: modal component to hold our mpr warning and user verification form.
* **Organization-user.module/oss.module**: New module for org user actions
* **organization-options**: move mpr enrollment from this class to a separate component.
* **messages.json**: Now that it's a modal with interaction, Are you sure was redundant.

## Screenshots

image (4)

## Before you submit

# Draft Reason
* [ ] Requires bitwarden/jslib#810


kspearrin:
@MGibson1 Someone else closer to the new module/architecture work for web vault should probably look at this one rather than me.

src/app/modules/organizations/users/enroll-master-password-reset.component.html
```diff
@@ -0,0 +1,59 @@
...
+          appA11yTitle="{{ 'close' | i18n }}"
+        >
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+        <app-callout type="warning" *ngIf="!isEnrolled">
+          {{ "resetPasswordEnrollmentWarning" | i18n }}
+        </app-callout>
+        <app-user-verification [(ngModel)]="verification" ngDefaultControl name="secret">
```
eliykat, line 32: Even though it's used this way elsewhere, I don't actually think we should use `ngDefaultControl` with this component. `ngDefaultControl` uses the default Control Value Accessor (CVA), whereas `UserVerificationComponent` implements its own CVA interface.

src/app/modules/organizations/users/enroll-master-password-reset.component.html
```diff
@@ -0,0 +1,59 @@
...
+          <i
+            class="bwi bwi-spinner bwi-spin"
+            title="{{ 'loading' | i18n }}"
+            *ngIf="form.loading"
+          ></i>
+          <span>
+            {{ "submit" | i18n }}
+          </span>
+        </button>
+        <button
```
eliykat, line 46: Can you use the CL button here instead? (`bit-button`)

src/app/modules/vault-filter/components/organization-options.component.ts
```diff
@@ -22,6 +25,9 @@ export class OrganizationOptionsComponent {
   policies: Policy[];
   loaded = false;
 
+  @ViewChild("enrollResetPasswordTemplate", { read: ViewContainerRef, static: true })
```
eliykat, line 28: Is this used anywhere?

src/app/modules/vault-filter/components/organization-options.component.ts
```diff
@@ -113,70 +120,12 @@ export class OrganizationOptionsComponent {
...
-      await this.load();
-    } catch (e) {
-      this.platformUtilsService.showToast("error", this.i18nService.t("errorOccurred"), e.message);
-      this.logService.error(e);
-    }
+    const ref = this.modalService.open(EnrollMasterPasswordReset, {
+      allowMultipleModals: true,
+      data: {
+        organization: org,
+        onSuccess: () => ref.close(),
```
eliykat, line 127: This doesn't seem like the right way to close the modal. Compare to `PasswordRepromptComponent` - it injects the `ModalRef` instead: https://github.com/bitwarden/jslib/blob/master/angular/src/components/password-reprompt.component.ts#L19

src/app/modules/organizations/users/enroll-master-password-reset.component.ts
```diff
@@ -0,0 +1,101 @@
...
+      await this.formPromise;
+      this.onSuccess();
+      this.platformUtilsService.showToast("success", null, this.i18nService.t(toastStringRef));
+    } catch (e) {
+      this.logService.error(e);
+    }
+  }
+
+  close() {
+    this.close();
```
eliykat, line 95: What circular business is this?

eliykat (changes requested):

eliykat:
Forgot to mention - I'm not sure that creating an `OrganizationUsersModule` is the right way to organize it, but I'm not sure where else it would fit given that it's just linked to from the vault page. So it's fine for now, but you might be interested in the discussion here: https://github.com/bitwarden/web/pull/1661

src/app/modules/vault-filter/components/organization-options.component.ts
Hinton: Yes, the `modalRef` handles closing and sending back potential responses on close.

src/app/modules/organizations/users/enroll-master-password-reset.component.ts
MGibson1: lol don't look at that!

MGibson1:
I've seen the discussion there and internally, but it seems that discussion has stagnated. We have comments discouraging adding to our existing bucket, so I'm kind of left with little recourse if my component doesn't fit well anywhere 🤷

eliykat (approved):
Yes, fair point about not having any other module to put it in. I'll try to restart those discussions soon so we can tame our structure a little.

addisonbeck:
I think it's fine to merge this as is, but in general the goal with modules is for them to line up with one of [these](https://angular.io/guide/module-types#summary-of-ngmodule-categories) types. It could be argued that `OrganizationUser` is a domain, but I would say Bitwarden's current domains are `Password Management` and `Secure File/Text Sharing` and that anything lower level then that is just a data model that applies to one of those domains. 

With that in mind, I would just call this `EnrollMasterPasswordResetModule` and drop the affiliation to the data model `OrganizationUser`. That's similar feedback to what is going on with the in-flight `OrganizationsModule`, but the changes aren't dependent on each other.

differsthecat (approved):

PR #1687, opened 2022-05-17
[SG-295] Use admin collections endpoint for org vault

eliykat:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Tech debt (refactoring, code cleanup, dependency upgrades, etc)
- [ ] Build/deploy pipeline (DevOps)
- [ ] Other

## Objective

See https://github.com/bitwarden/jslib/pull/806.

**This will be picked to rc**

## Code changes

* **src/app/modules/vault/modules/organization-vault/organization-vault.component.ts** - set the new property on `VaultFilter`, so `VaultFilterService` knows to fetch from the admin endpoint
* **src/app/modules/vault-filter/vault-filter.component.ts** - delete unused method (see jslib pr)
* **src/app/modules/vault-filter/vault-filter.module.ts** - the VaultFilterService uses the `Injectable()` decorator, so we don't need to specify dependencies manually.
* **src/app/modules/vault-filter/vault-filter.service.ts** - deleted. This was never used.

## Before you submit

- [x] I have checked for **linting** errors (`npm run lint`) (required)
- [ ] This change requires a **documentation update** (notify the documentation team)
- [ ] This change has particular **deployment requirements** (notify the DevOps team)

PR #1700, opened 2022-05-20
[fix] Pull org admin vault collections directly from the API

addisonbeck:
https://bitwarden.atlassian.net/browse/SG-295

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Continue the work done by @eliykat on https://github.com/bitwarden/web/pull/1687 and https://github.com/bitwarden/jslib/pull/806 with a slightly different direction.

Address a bug where creating a collection and then navigating to the organization vault does not show the new collection when creating items.

## Code changes
Along the way of the `GroupingsComponent -> VaultFiltersModule` restructure for the End User Vault Refresh initiative we lost the distinction between an organization user using the organization vault who can manage all collections from an organization user using the organization vault that can not manage all collections. If someone can manage all collections we are supposed to be grabbing those from the API, but instead we are using local storage.

To fix this I have:
1. Extended the `VaultFiltersService` to account for the web-specific use case of the Organization Vault, and added a `buildAdminCollections` method to it.
1. Created an `OrganizationVaultFiltersComponent` (it seemed like it was time to do this) that extends the `VaultFiltersComponent`.
2. Moved the existing Organization Vault specific logic out of `VaultFiltersComponent` and into `OrganizationVaultFiltersComponent` (`organization`, `initCollections()`).
3. Inform the `OrganizationVaultFiltersComponent.initCollections()` method of the admin use case that should gather collections from the API.
4. Inform the `OrganizationVaultFiltersComponent.reloadCollectionsAndFolders()` method that it doesn't need to worry about folders at all, and should reload collections with the Admin use case considered. There is probably still some room to clean this up, namely separating `reloadCollections` from `reloadFolders` since `reloadCollectionsAndFolders` doesn't apply to the `OrganizationVault`, but I was avoiding touching jslib with this change.

Another bug that these changes introduced are that we are building the initial collection list twice in the organization vault: once on `ngOnInit` in the `VaultFiltersComponent` and once from the query parameter subscription in `OrganizationVaultComponent`. I hacked around this for now with a comment on how to better deal with it later during slower times.

I also removed the `deps` array from the `VaultFilterService` provider in `VaultFiltersModule`, following the same action from @eliykat .

## Screenshots
https://user-images.githubusercontent.com/15897251/169573650-68fcf4fa-d598-4425-b431-14caf7538203.mov

## Before you submit


src/app/modules/vault/modules/organization-vault/organization-vault.component.html
```diff
@@ -4,15 +4,15 @@
       <div class="groupings">
         <div class="content">
           <div class="inner-content">
-            <app-vault-filter
+            <app-organization-vault-filter
               #vaultFilter
               [showFolders]="false"
```
differsthecat, line 9: The `showFolders` input was added for this case (don't show it for orgs). Can we do something like default it to `false` in the org vault filter component? Then we don't have to pass it and that cleans this up a little.

src/app/modules/vault/modules/organization-vault/organization-vault.component.ts
```diff
@@ -57,6 +58,11 @@ export class OrganizationVaultComponent implements OnInit, OnDestroy {
   trashCleanupWarning: string = null;
   activeFilter: VaultFilter = new VaultFilter();
 
+  // This is a hack to avoid redundant api calls that fetch OrganizationVaultFilterComponent collections
+  // When it makes sense to do so we should leverage some other communication method for change events that isn't directly tied to the query param for organizationId
+  // i.e. exposing the VaultFiltersService to the OrganizationSwitcherComponent to make relevant updates from a change event instead of just depending on the router
+  skipFilterCollectionRefresh = true;
```
eliykat, line 64: I suggest we rename this `firstLoaded`, which is the naming we use elsewhere for this pattern. It communicates better how the flag works.

I think this is an acceptable short-term fix, but I think it does mean we should try to implement your refactor soon.

eliykat (changes requested):
Thanks for your work on this! Interesting to see another way to approach the problem.

src/app/modules/vault/modules/organization-vault/organization-vault.component.ts
eliykat: Also worth noting this issue is also present for the individual vault - it hits local data, so it's less important, but ideally any long-term fix here should apply to both vaults.

src/app/modules/vault-filter/vault-filter.component.ts
```diff
@@ -18,17 +18,11 @@ export class VaultFilterComponent extends BaseVaultFilterComponent {
   searchPlaceholder: string;
   searchText = "";
 
-  organization: Organization;
-
-  constructor(vaultFilterService: VaultFilterService) {
+  constructor(protected vaultFilterService: VaultFilterService) {
```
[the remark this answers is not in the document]
addisonbeck: Thoughts on https://github.com/bitwarden/jslib/pull/822 and [[refactor] Establish pattern for interfaces in module scoped services](https://github.com/bitwarden/web/pull/1700/commits/c808aa8a15be1ec121754d565ab6aac910dbbe74)?

Pros:
- Removes the constructor in `web/VaultFilterComponent`, uses DI to inject the right service from an abstraction
- Still allows for extending module services and their abstractions in clients for stuff that doesn't belong in jslib so we can override methods or make new ones

Cons:
- We have to do the weird thing where we provide the right service for an extended interface and its base interface
- We still have to have a special constructor in `web/OrganizationVaultFiltersComponent` so it knows about the web specific `buildAdminCollections` method. I think this is less confusing, since if you remove the constructor you will immediately get a syntax error.

src/app/modules/vault-filter/vault-filter.service.ts
```diff
@@ -1,3 +1,54 @@
...
+import { StateService } from "jslib-common/abstractions/state.service";
+import { CollectionData } from "jslib-common/models/data/collectionData";
+import { Collection } from "jslib-common/models/domain/collection";
+import { CollectionDetailsResponse } from "jslib-common/models/response/collectionResponse";
+import { CollectionView } from "jslib-common/models/view/collectionView";
+
+@Injectable()
+export class VaultFilterService extends BaseVaultFilterService {
+  constructor(
+    protected stateService: StateService,
```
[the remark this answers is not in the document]
addisonbeck: [[fix] Remove access modifiers from constructor params for an extended…](https://github.com/bitwarden/web/pull/1700/commits/3c0d78c61cace8051e9137bc2b6317262e27e37d)

src/app/modules/vault/modules/organization-vault/organization-vault.component.html
addisonbeck: Fixed with [[fix] Simplify hiding unwanted filters in the org vault](https://github.com/bitwarden/web/pull/1700/commits/e07deff505d6c3b2b57c3f8121f8908a3581b069)

I tweaked this a bit and also removed these properties completely, similar inputs already existed in jslib.

differsthecat: Awesome, yeah that's much better!

Hinton:
Closing due to move to mono repo

addisonbeck:
Reopening since it looks like we are hotfixing this after all

src/app/modules/vault-filter/vault-filter.component.ts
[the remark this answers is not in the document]
eliykat: Marking this as resolved because it's out of scope for this PR

src/app/modules/vault-filter/vault-filter.component.ts
```diff
@@ -1,34 +1,24 @@
...
-
   @Output() onSearchTextChanged = new EventEmitter<string>();
 
   searchPlaceholder: string;
   searchText = "";
 
-  organization: Organization;
-
-  constructor(vaultFilterService: VaultFilterService) {
+  constructor(protected vaultFilterService: VaultFilterService) {
```
eliykat, line 17:
```suggestion
  constructor(protected vaultFilterService: VaultFilterService) {
  // This empty constructor is required to provide the web vaultFilterService subclass to super()
  // TODO: refactor this to use proper dependency injection
```

src/app/modules/vault-filter/organization-vault-filter.component.ts
```diff
@@ -0,0 +1,28 @@
...
+import { VaultFilterComponent } from "./vault-filter.component";
+
+@Component({
+  selector: "app-organization-vault-filter",
+  templateUrl: "vault-filter.component.html",
+})
+export class OrganizationVaultFilterComponent extends VaultFilterComponent {
+  @Input() hideOrganizations = true;
+  @Input() hideFavorites = true;
+  @Input() hideFolders = true;
```
eliykat, line 14: We should remove the `@Input()` decorator on these declarations, as we never expect to use them as inputs in this subclass. They're effectively just constants used in the template.

```suggestion
  hideOrganizations = true;
  hideFavorites = true;
  hideFolders = true;
```

eliykat (changes requested):
This is fine for a hotfix, just left a couple of comments. Assuming we miss each other, please feel free to get someone else to give an approving review once these are addressed.

One other thing: Did you intend to bump jslib? Those changes seem unrelated. (and shouldn't be included for a hotfix)

eliykat:
PS. don't forget to cherry-pick to `clients` after merging :D

addisonbeck:
> One other thing: Did you intend to bump jslib? Those changes seem unrelated. (and shouldn't be included for a hotfix)

This happened during a merge. The release branch will be manually targeting an `rc` branch for jslib, so those changes will not be included.

differsthecat (approved):

PR #1706, opened 2022-05-25
Revert "[ps-136] Igonre accented characters in vault search"

gbubemismith:
Reverts bitwarden/web#1690. 

The logic to remove accents from `searchText` has been moved to the `search service` in https://github.com/bitwarden/jslib/pull/812. This PR is no longer needed.


eliykat:
The normalization logic was moved into `searchService`, so it's no longer needed in the component (before calling `searchService`).

eliykat (approved):

